PR #16, opened 2025-09-18
Add mcp server

haixuanTao:
This is a quick rewrite of our existing openai realtime script as an mcp server that can be configured with:

```json
{
  "servers": {
    "reachy": {
      "command": "uv",
      "args": [
        "--directory",
        "/path/to/reachy_mini_conversation_demo",
        "run",
        "reachy-mini-mcp"
      ],
    }
  }
}
```

And then let the llm mcp tool call Reachy Mini.  :) 

I have reduced the number of tools in order for lower latency on local llm but we can add more in the future!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,9 @@ dependencies = [
     "dotenv",
     "ultralytics",
     "supervision",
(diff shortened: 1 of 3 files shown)


alozowski:
Closing this PR as it is outdated

PR #19, opened 2025-09-18
18 graceful quit

FabienDanieau:
Exiting the app should be cleaner

### DIFF
--- a/src/reachy_mini_conversation_demo/main.py
+++ b/src/reachy_mini_conversation_demo/main.py
@@ -1,7 +1,6 @@
 """Entrypoint for the Reachy Mini conversation demo."""
 
 import os
-from threading import Thread
 
 import gradio as gr
 from fastapi import FastAPI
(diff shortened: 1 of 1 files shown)


src/reachy_mini_conversation_demo/audio/head_wobbler.py
```diff
@@ -54,7 +54,13 @@ async def enable(self, stop_event: asyncio.Event) -> None:
         self._consumer_loop = loop
 
         while not stop_event.is_set():
-            sr, chunk = await self.audio_queue.get()  # (1,N) int16
+            try:
+                sr, chunk = self.audio_queue.get_nowait()  # (1,N) int16
+            except QueueEmpty:
+                # avoid while to never exit
+                await asyncio.sleep(0.02)
```
RemiFabre, line 61: I'll let the magic number slide because the sleep duration calculation here is overly complex (might even be the cause of a problem I'm investigating, I might rework it)

RemiFabre: Using nowait seems to be a better idea in an asyncio context tho, well done

Issue #23, opened 2025-09-20
remove asyncio - use threads


FabienDanieau:
#22 has performance issue. Let's try the other solution

PR #24, opened 2025-09-20
enhancement #23: remove asyncio except from openai handler

FabienDanieau:
replace #22 . asyncio cannot guarantee the 50Hz loop

### DIFF
--- a/src/reachy_mini_conversation_demo/audio/head_wobbler.py
+++ b/src/reachy_mini_conversation_demo/audio/head_wobbler.py
@@ -38,6 +38,7 @@ class HeadWobbler:
 
     def start(self) -> None:
         """Start the head wobbler loop in a thread."""
+        self._stop_event.clear()
         self._thread = threading.Thread(target=self.working_loop, daemon=True)
         self._thread.start()
         logger.info("Head wobbler started")


src/reachy_mini_conversation_demo/moves.py
```diff
@@ -395,17 +398,30 @@ def set_neutral(self) -> None:
...
         Single set_target() call with pose fusion.
         """
-        logger.info("Starting enhanced movement control loop (50Hz)")
+        logger.debug("Starting enhanced movement control loop (50Hz)")
 
         loop_count = 0
         last_print_time = time.time()
 
-        while not stop_event.is_set():
+        while not self._stop_event.is_set():
```
RemiFabre, line 429: _stop_event is never cleared no? Should be done in the start function (even though the threads are currently not meant to be restarted)

src/reachy_mini_conversation_demo/camera_worker.py
```diff
@@ -80,26 +81,39 @@ def set_head_tracking_enabled(self, enabled: bool) -> None:
...
-    async def enable(self, stop_event: asyncio.Event) -> None:
+    def start(self) -> None:
+        """Start the camera worker loop in a thread."""
+        self._thread = threading.Thread(target=self.working_loop, daemon=True)
+        self._thread.start()
+        logger.info("Camera worker started")
+
+    def stop(self) -> None:
+        """Stop the camera worker loop."""
+        self._stop_event.set()
```
RemiFabre, line 93: _stop_event is never cleared no? Should be done in the start function (even though the threads are currently not meant to be restarted)

FabienDanieau: true I'll add a clear to each start function
Edit: you did it perfect

PR #57, opened 2025-10-14
Add github workflows

alozowski:
To be able to use github automation I had to switch from installing fastrtc from github to a pypi version, thus this PR brings back the verbose logs output, because [this commit](https://github.com/gradio-app/fastrtc/commit/27f054934f15070750ee0383a94c103231bd0bbb) wasn't included into the pypi release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,3 @@ ignore_missing_imports = true
 strict = true
 show_error_codes = true
 warn_unused_ignores = true
-
-[[tool.mypy.overrides]]
-module = "reachy_mini_conversation_demo.*"
-disable_error_code = ["misc"]
--- a/src/reachy_mini_conversation_demo/moves.py
+++ b/src/reachy_mini_conversation_demo/moves.py
@@ -61,7 +61,7 @@ CONTROL_LOOP_FREQUENCY_HZ = 100.0  # Hz - Target frequency for the movement cont
 FullBodyPose = Tuple[NDArray[np.float32], Tuple[float, float], float]  # (head_pose_4x4, antennas, body_yaw)
 
 
-class BreathingMove(Move):
+class BreathingMove(Move):  # type: ignore
     """Breathing move with interpolation to neutral and then continuous breathing patterns."""
 
     def __init__(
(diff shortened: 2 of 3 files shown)


src/reachy_mini_conversation_demo/moves.py
```diff
@@ -57,15 +58,15 @@
 CONTROL_LOOP_FREQUENCY_HZ = 100.0  # Hz - Target frequency for the movement control loop
 
 # Type definitions
-FullBodyPose = Tuple[np.ndarray, Tuple[float, float], float]  # (head_pose_4x4, antennas, body_yaw)
+FullBodyPose = Tuple[NDArray[np.float32], Tuple[float, float], float]  # (head_pose_4x4, antennas, body_yaw)
 
 
-class BreathingMove(Move):
+class BreathingMove(Move):  # type: ignore[misc]
```
FabienDanieau: This ignore shouldn't be there (it's ok with other classes). It may be something else.
Like the evaluate has not the same signature than the abstract class.

pyproject.toml
```diff
@@ -40,7 +40,11 @@ all_vision = [
 ]
 
 [dependency-groups]
-dev = ["pytest", "ruff==0.12.0"]
+dev = [
+  "pytest",
+  "ruff==0.12.0",
+  "mypy>=1.18.2",
```
FabienDanieau: Use == since different versions can lead to different errors. We want to make sure that the user is using the same as the CI

src/reachy_mini_conversation_demo/vision/processors.py
```diff
@@ -89,11 +90,11 @@ def initialize(self) -> bool:
 
     def process_image(
         self,
-        cv2_image: np.ndarray,
+        cv2_image: np.ndarray[Any, Any],
```
FabienDanieau: remaining np.ndarray -> NDArray

src/reachy_mini_conversation_demo/moves.py
FabienDanieau: same comments for other classes except the RecordedMove

Issue #73, opened 2025-10-24
Use non-beta version of realtime


andimarafioti:
In this line: https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/27b505e0e969ae68cd97f81f663cf1ccff65a0e7/src/reachy_mini_conversation_app/openai_realtime.py#L52
we are using the beta version of realtime. Changing it is more than just that line since the API changed. We should get this done before OpenAI completely sunsets this :)

PR #75, opened 2025-10-24
enhancement #76: Migrate from beta OpenAI to GA

alozowski:
This PR fixes the following:
- Migrate from OpenAI Realtime Beta to GA – ensures full compatibility with the GA Realtime API, [see this issue](https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/73)

### DIFF
--- a/src/reachy_mini_conversation_app/openai_realtime.py
+++ b/src/reachy_mini_conversation_app/openai_realtime.py
@@ -132,9 +132,6 @@ class OpenaiRealtimeHandler(AsyncStreamHandler):
                     "response.completed",             # text-only completion
                 ):
                     logger.debug("response completed")
-                    # TODO: remove later
-                    # if self.deps.head_wobbler is not None:
-                    #     self.deps.head_wobbler.reset()
 
                 if event.type == "response.created":
                     logger.debug("Response created")


andimarafioti:
https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/73

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -46,26 +53,58 @@ def copy(self) -> "OpenaiRealtimeHandler":
...
-                    "input_audio_transcription": {
-                        "model": "whisper-1",
-                        "language": "en",
+        async with self.client.realtime.connect(model=config.MODEL_NAME) as conn:
+            try:
+                await conn.session.update(
+                    session={
+                        "type": "realtime",
+                        "instructions": SESSION_INSTRUCTIONS,
+                        "audio": {
```
andimarafioti, line 79: did you lose the turn detection type from beta to GA? Even if the default is still server vad, it would be good to keep it explicit

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -35,6 +35,13 @@ def __init__(self, deps: ToolDependencies):
         )
         self.deps = deps
 
+        # Add resampling ratio
+        self.target_input_rate = 24000  # OpenAI requirement
+        self.resample_ratio = self.target_input_rate / self.input_sample_rate
+
+        # Buffer for accumulating partial samples during resampling
+        self.resample_buffer = np.array([], dtype=np.int16)
+
```
andimarafioti: vibe coded variable not used (how did mypy miss this xD)

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -187,14 +237,14 @@ async def start_up(self) -> None:
                                 ),
                             )
 
-                    if not self.is_idle_tool_call:
+                    if self.is_idle_tool_call:
+                        self.is_idle_tool_call = False
+                    else:
```
andimarafioti, line 249: Maybe a comment explaining that this comes from the idle movement? I was confused by this

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -249,21 +312,39 @@ async def send_idle_signal(self, idle_duration: float) -> None:
...
+                response={
+                    "instructions": (
+                        "You MUST respond with function calls only—no speech or text. "
+                        "Pick exactly one of these tools and call it with sensible args: "
+                        + ", ".join(allowed)
+                        + ". Prefer small, varied motions; occasionally choose 'do_nothing'."
+                    ),
+                    "tool_choice": "required",
+                },
+            )
```
andimarafioti: is this related to the GA? It was working well for me before

andimarafioti (approved):
Good work! I don't see anything too big to change, but I left a few comments.
A bit of sanity, I would not push the uv.locks, especially if there was no change to the dependencies.

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: I've been trying to force the tool calling but reverting this, the initial approach works well

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: I removed it because it's default, yes, but let's restore like this to make it explicit:
```
"input": {
    "format": {
        "type": "audio/pcm",
        "rate": self.target_input_rate,
    },
    "transcription": {
        "model": "whisper-1",
        "language": "en"
    },
    "turn_detection": {
        "type": "server_vad",
    },
},
```

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: Added the explanation! 👍🏻

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -81,13 +120,17 @@ async def start_up(self) -> None:
...
+                    "response.audio.done",            # GA
+                    "response.output_audio.done",     # GA alias
+                    "response.audio.completed",       # legacy (for safety)
+                    "response.completed",             # text-only completion
+                ):
                     logger.debug("response completed")
-                    if self.deps.head_wobbler is not None:
-                        self.deps.head_wobbler.reset()
+                    # if self.deps.head_wobbler is not None:
+                    #     self.deps.head_wobbler.reset()
```
alozowski: If we comment out `head_wobbler.reset()` here, if fixes the movement error, and the robot moves continuously throughout the entire speech response – to verify if it's the correct approach

alozowski:
Removed `uv.lock` file changes from this PR to keep it clean

src/reachy_mini_conversation_app/prompts.py
haixuanTao, line 1: Although fine for this PR, I don't think we should add system prompts change within PR otherwise it might introduce all kinds of merge conflicts no?

alozowski: Agree, removing the prompt change

alozowski:
@andimarafioti thank you for the error! I think it was a hiccup on the OpenAI's side – I will add a connection retry in a next PR

RemiFabre:
Hello, tested this version for about 30 minutes doing the usual drills. As far as I can tell, the code works as it should!
I'm not sure why a change in the wobble reset call was needed and what the bug was, but let keep this as is as I was unable to find any weird behavior.

I have observed some more general weird/annoying things. Everyday it seems the behavior changes depending on the server load at openAI. Yesterday for example was a nightmare as consecutive tool calls (like "do A, B then C)" were executed with several seconds of lag BETWEEN the tool calls.
Today the model was very "dumb". I'd describe a game like "when I say GO, tak a picture, analyze the emotion of the person in the center and play the closest emotion you've got". It works 1 out of 5 times. Sometimes it tells me about the emotion without playing the emotion, sometimes I say go, it takes a picture and doesn't do anything, after a few exchanges it completely forgets about the game.

I've observed this variability of performance many times and it's a nightmare to debug our code :/

PR #79, opened 2025-10-27
Flexible demo system to customize prompts and tools, and share them in reusable libraries, staying out of the core components.

scienceetonnante:
Implements a flexible demo system that allows custom personalities and tools to be loaded via environment variable, and shared in librairies for re-use, while maintaining minimal intrusion to the existing codebase.

### Core Implementation
- **prompts.py**: Added `get_session_instructions()` to dynamically load instructions from the specified demo
- **tools.py**: Added `_load_demo_tools()` to auto-discover and load tools from the specified demo
- **openai_realtime.py**: Updated to use `get_session_instructions()` instead of static `SESSION_INSTRUCTIONS` (still used as a default when no demo is specified).

### Structure
- Created `src/demos/` package for modular use cases
- Created `src/prompts_library/` package for re-using prompt pieces
- Created `src/tools_library/` package for re-using tools
- Each demo is a package with `__init__.py` containing:
  - `instructions: str` - Custom personality/behavior instructions
  - Tool classes - Custom tools (auto-discovered via inheritance)

This way, creating a demo can be done entirely without modifying files in the `src/reachy_mini_conversation_app/` folder.

### Example Use Case
- **src/demos/stone/**: Example "stone-throwing robot" demo
  - Starts from a shared prompt piece from the prompts library (witty_identity)
  - Adds demo-specific instructions
  - Define a custom `ThrowStone` tool
  - Adds a tool from the shared tools library (sweep_look)

To define the demo to be loaded
```
# In .env
DEMO="stone"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Create custom profiles with dedicated instructions and enabled tools!
 
 Set `REACHY_MINI_CUSTOM_PROFILE=<name>` to load `src/reachy_mini_conversation_app/profiles/<name>/` (see `.env.example`). If unset, the `default` profile is used.
 
-Each profile needs two files: `instructions.txt` (prompt text) and `tools.txt` (list of allowed tools).
+Each profile requires two files: `instructions.txt` (prompt text) and `tools.txt` (list of allowed tools), and optionally contains custom tools implementations.
 
 ### Custom instructions
 Write plain-text prompts in `instructions.txt`. To reuse shared prompt pieces, add lines like:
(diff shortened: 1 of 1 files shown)


src/reachy_mini_conversation_app/prompts/identities/witty_identity.txt
haixuanTao, line 1: I wonder if we shouldn't use text file or json for system prompt so that's it easier for people with no python bagage to modify. It might also make things like, escaping those files on git easier.

I had a person on the stand that did not understand python syntax couldn't figure out how to make it work.

scienceetonnante: Good point ! I'll change this, thanks

RemiFabre (approved):

Issue #82, opened 2025-10-29
Handle transient ConnectionClosedError in Realtime WebSocket


alozowski:
Occasionally, the app crashes with:
```
websockets.exceptions.ConnectionClosedError: no close frame received or sent
```
See here: https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/75#issuecomment-3460941793

This seems to happen when the OpenAI Realtime server or network drops the connection unexpectedly.
We should handle this gracefully: catch the exception, log a warning, and optionally retry or shut down cleanly instead of crashing the process.

PR #98, opened 2025-11-12
Fix ConnectionClosedError

alozowski:
Fix [Issue 82](https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/82)

The `start_up()` function was modified to include a retry loop that catches `ConnectionClosedError`, performs a short backoff, and retries the realtime connection up to three times.
It also now guarantees cleanup by resetting `self.connection` in a `finally` block after any abrupt websocket closure.

The `shutdown()` function was also updated to safely handle already-closed connections by catching `ConnectionClosedError` and other close-related exceptions, ensuring graceful termination without raising errors.

I also added a related test, but I'm bad at them and I would appreciate the review.

### DIFF
--- a/src/reachy_mini_conversation_app/openai_realtime.py
+++ b/src/reachy_mini_conversation_app/openai_realtime.py
@@ -76,221 +76,227 @@ class OpenaiRealtimeHandler(AsyncStreamHandler):
         max_attempts = 3
         for attempt in range(1, max_attempts + 1):
             try:
-                async with self.client.realtime.connect(model=config.MODEL_NAME) as conn:
-                    try:
-                        await conn.session.update(
-                            session={
-                                "type": "realtime",
-                                "instructions": SESSION_INSTRUCTIONS,
-                                "audio": {
-                                    "input": {
-                                        "format": {
-                                            "type": "audio/pcm",
-                                            "rate": self.target_input_rate,
-                                        },
-                                        "transcription": {
-                                            "model": "whisper-1",
-                                            "language": "en"
-                                        },
-                                        "turn_detection": {
-                                            "type": "server_vad",
-                                            "interrupt_response": True,
-                                        },
-                                    },
-                                    "output": {
-                                        "format": {
-                                            "type": "audio/pcm",
-                                            "rate": self.output_sample_rate,
-                                        },
-                                        "voice": "cedar",
-                                    },
-                                },
-                                "tools": ALL_TOOL_SPECS,  # type: ignore[typeddict-item]
-                                "tool_choice": "auto",
-                            },
-                        )
-                    except Exception:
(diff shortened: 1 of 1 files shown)


src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -68,206 +70,227 @@ def resample_audio(self, audio: NDArray[np.int16]) -> NDArray[np.int16]:
         return cast(NDArray[np.int16], resampled.astype(np.int16))
 
     async def start_up(self) -> None:
-        """Start the handler."""
+        """Start the handler with minimal retries on unexpected websocket closure."""
```
andimarafioti, line 73: This start_up method is already big enough as it is, could you wrap the actual start_up of the client in another method and have this loop with tries be the only thing that's exposed on start_up?

Something like

for attempt in max_retries:
  try:
    create_client
    return
  except error:
    handle

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -68,206 +70,227 @@ def resample_audio(self, audio: NDArray[np.int16]) -> NDArray[np.int16]:
...
+                    except ConnectionClosedError as e:
+                        # Abrupt close (e.g., "no close frame received or sent") → retry
+                        logger.warning(
+                            "Realtime websocket closed unexpectedly (attempt %d/%d): %s",
+                            attempt, max_attempts, e
+                        )
+                        if attempt < max_attempts:
+                            # small jittered backoff
+                            await asyncio.sleep(1.0 + random.uniform(0, 0.5))
+                            continue
```
andimarafioti: For connection errors, you usually do exponentially increasing wait times. So the first might be 1s, then 4s and finally 16s or similar. It's ok to add some jitter, as you did, but also give some time for the server to recover.

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -305,8 +328,14 @@ async def emit(self) -> Tuple[int, NDArray[np.int16]] | AdditionalOutputs | None
     async def shutdown(self) -> None:
         """Shutdown the handler."""
         if self.connection:
-            await self.connection.close()
-            self.connection = None
+            try:
+                await self.connection.close()
+            except ConnectionClosedError:
+                pass
```
andimarafioti: I woudn't just pass here, you can use the logger.debug as well

andimarafioti (changes requested):
Good work! I left some suggestions for improvements, mainly to make this more clear since the method keeps on growing, let me know what you think!

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: Completely agree!

PR #103, opened 2025-11-16
Fix delay caused by self.sway.feed in HeadWobbler

CarolinePascal:
This PR simply puts the `self_base_ts` initialization before the `self.sway.feed` call to deduce the sway computation duration from the target time. 
This is particularly relevant when the received audio sample is quite long.


RemiFabre:
Good catch! Merging this, but I'm afraid there is something else creating the rare cases of ~1s desync and we have not found a repeatable test for this. 
Let's test this version to see if this fixes it

PR #122, opened 2025-12-01
62 appify the demo

apirrone:
Add the necessary code so that this python package can be installed and ran as a Reachy Mini App from the Dashboard. 
Shouldn't break anything that worked before

### DIFF
--- a/src/reachy_mini_conversation_app/console.py
+++ b/src/reachy_mini_conversation_app/console.py
@@ -285,21 +285,18 @@ class LocalStream:
             # Try to validate by checking if we can fetch the models
             try:
                 import httpx
-                headers = {
-                    "Authorization": f"Bearer {key}",
-                    "Content-Type": "application/json"
-                }
+
+                headers = {"Authorization": f"Bearer {key}", "Content-Type": "application/json"}
                 async with httpx.AsyncClient(timeout=10.0) as client:
-                    response = await client.get(
-                        "https://api.openai.com/v1/models",
-                        headers=headers
-                    )
+                    response = await client.get("https://api.openai.com/v1/models", headers=headers)
                     if response.status_code == 200:
                         return JSONResponse({"valid": True})
                     elif response.status_code == 401:
                         return JSONResponse({"valid": False, "error": "invalid_api_key"}, status_code=401)
                     else:
-                        return JSONResponse({"valid": False, "error": "validation_failed"}, status_code=response.status_code)
+                        return JSONResponse(
+                            {"valid": False, "error": "validation_failed"}, status_code=response.status_code
+                        )
             except Exception as e:
                 logger.warning(f"API key validation failed: {e}")
                 return JSONResponse({"valid": False, "error": "validation_error"}, status_code=500)
@@ -321,6 +318,7 @@ class LocalStream:
         if self._instance_path:
             try:
                 from dotenv import load_dotenv
+
                 from reachy_mini_conversation_app.config import set_custom_profile
 
                 env_path = Path(self._instance_path) / ".env"


src/reachy_mini_conversation_app/main.py
```diff
@@ -27,17 +26,28 @@ def update_chatbot(chatbot: List[Dict[str, Any]], response: Dict[str, Any]) -> L
...
-def main() -> None:
+def main():
     """Entrypoint for the Reachy Mini conversation app."""
     args = parse_args()
+    run(args)
+
+
+def run(args, robot=None, app_stop_event: threading.Event = None):
+    """Run the Reachy Mini conversation app."""
+    from reachy_mini_conversation_app.moves import MovementManager
```
apirrone, line 45: Putting these dependencies here makes the dashboard faster to load when the conversation app is installed

src/reachy_mini_conversation_app/main.py
```diff
@@ -27,17 +26,28 @@ def update_chatbot(chatbot: List[Dict[str, Any]], response: Dict[str, Any]) -> L
...
+    from reachy_mini_conversation_app.audio.head_wobbler import HeadWobbler
 
     logger = setup_logger(args.debug)
     logger.info("Starting Reachy Mini Conversation App")
 
     if args.no_camera and args.head_tracker is not None:
         logger.warning("Head tracking is not activated due to --no-camera.")
 
-    robot = ReachyMini()
+    if robot is None:
```
apirrone, line 57: When running as an dashboard app, the dashboard provides the reachy_mini instance

src/reachy_mini_conversation_app/main.py
```diff
@@ -103,15 +118,24 @@ def main() -> None:
     if vision_manager:
         vision_manager.start()
 
+    def poll_stop_event():
+        """Poll the stop event to allow graceful shutdown."""
```
apirrone, line 168: Handle the stop event provided by the dashboard

src/reachy_mini_conversation_app/main.py
```diff
@@ -121,8 +145,25 @@ def main() -> None:
 
         # prevent connection to keep alive some threads
         robot.client.disconnect()
+        time.sleep(1)
         logger.info("Shutdown complete.")
 
 
+class ReachyMiniConversationApp(ReachyMiniApp):
```
apirrone: The class that is loaded by the dashboard

andimarafioti:
passing `--gradio` works, so maybe this was here for a while and people didn't notice?

README.md
```diff
@@ -1,3 +1,15 @@
+---
+title: Reachy Mini Conversation App
+emoji: 🎤
+colorFrom: red
+colorTo: blue
+sdk: static
```
alozowski, line 6: are we sure that `sdk: static` is the correct solution here? should it be `sdk: gradio` instead? I would verify the HF documentation

src/reachy_mini_conversation_app/main.py
andimarafioti: Could you add this comment to the code? Otherwise the next person might move them again

src/reachy_mini_conversation_app/main.py
```diff
@@ -121,8 +147,25 @@ def main() -> None:
...
+class ReachyMiniConversationApp(ReachyMiniApp):  # type: ignore[misc]
+    """Reachy Mini Apps entry point for the conversation app."""
+
+    custom_app_url = "http://127.0.0.1:7860/"
+    dont_start_webserver = True
+
+    def run(self, reachy_mini: ReachyMini, stop_event: threading.Event) -> None:
+        """Run the Reachy Mini conversation app."""
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
```
alozowski, line 208: Gradio manages its own event loop. Creating one here causes multiple loops and task scheduling warnings. I tested and I can confirm it works without these lines, let's remove then

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -57,33 +57,50 @@ def __init__(self, deps: ToolDependencies):
...
             raise
 
     async def start_up(self) -> None:
         """Start the handler with minimal retries on unexpected websocket closure."""
-        self.client = AsyncOpenAI(api_key=config.OPENAI_API_KEY)
+        openai_api_key = config.OPENAI_API_KEY
+        if self.gradio_mode:
+            await self.wait_for_args()  # type: ignore[no-untyped-call]
+            args = list(self.latest_args)
+            textbox_api_key = args[3] if len(args[3]) > 0 else None
```
alozowski, line 156: Using `args[3]` assumes fastrtc always prepends exactly 2 internal args before `additional_inputs=[chatbot, api_key_textbox]`. This is fragile - if fastrtc changes its internal structure, this breaks. Also `len(args[3])` crashes if args < 4 elements

Can we use `args[-1] if len(args) >= 1 and args[-1] else None` instead? But it's still fragile, we need a more robust solution to extract an api key

src/reachy_mini_conversation_app/main.py
```diff
@@ -121,8 +147,25 @@ def main() -> None:
 
         # prevent connection to keep alive some threads
         robot.client.disconnect()
```
alozowski, line 194: `robot.client.disconnect()` runs even when robot is provided externally in `ReachyMiniConversationApp`. Should we only disconnect if `robot` was created locally by tracking ownership at the beginning of `run()`?

Use case: when switching apps through the dashboard, the dashboard launches conversation app with a robot instance, then user stops the app and line 149 disconnects the robot, then dashboard tries to launch another app with same robot but the robot is already disconnected, breaking the next app. Expected behavior is to only disconnect if we created the robot locally

This can be an issue with the appyfying since `ReachyMiniConversationApp.run()`now passes an external robot instance, which will be incorrectly disconnected

What do you think?

src/reachy_mini_conversation_app/main.py
```diff
@@ -75,16 +86,21 @@ def main() -> None:
     )
     logger.debug(f"Chatbot avatar images: {chatbot.avatar_images}")
 
-    handler = OpenaiRealtimeHandler(deps)
+    handler = OpenaiRealtimeHandler(deps, gradio_mode=args.gradio)
 
     stream_manager: gr.Blocks | LocalStream | None = None
 
     if args.gradio:
+        api_key_textbox = gr.Textbox(
```
alozowski, line 117: I would add placeholder here ([doc](https://www.gradio.app/docs/gradio/textbox)), something like:
```
api_key_textbox = gr.Textbox(
    label="OpenAI API Key",
    type="password",
    value=os.getenv("OPENAI_API_KEY") if not get_space() else "",
    placeholder="sk-..." if get_space() else "",
)
```

(I didn't start this code)

alozowski:
I also have these questions:
- should we validate the API key format before attempting to connect? it can go to config.py or so
- when I start `reachy-mini-conversation-app --gradio`, I can't use it even though I have an  `.env` file with the key – I still need to fill in the key form on the left. Can it be automatically pass from `.env` file if I have it?

src/reachy_mini_conversation_app/main.py
apirrone: We need to add this when running from the dashboard. We get this error if not : 

```
ERROR:reachy_mini.apps.manager.runner:Exception details: 'Traceback (most recent call last):
  File "/home/antoine/Pollen/reachy_mini/src/reachy_mini/apps/app.py", line 85, in wrapped_run
    self.run(reachy_mini, self.stop_event)
  File "/home/antoine/Pollen/reachy_mini_conversation_app/src/reachy_mini_conversation_app/main.py", line 171, in run
    run(args, robot=reachy_mini, app_stop_event=stop_event, settings_app=self.settings_app)
  File "/home/antoine/Pollen/reachy_mini_conversation_app/src/reachy_mini_conversation_app/main.py", line 89, in run
    handler = OpenaiRealtimeHandler(deps, gradio_mode=args.gradio)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/antoine/Pollen/reachy_mini_conversation_app/src/reachy_mini_conversation_app/openai_realtime.py", line 57, in __init__
    self.last_activity_time = asyncio.get_event_loop().time()
                              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/events.py", line 702, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'Thread-7 (wrapped_run)'.
````

src/reachy_mini_conversation_app/openai_realtime.py
apirrone: Yeah I agree that it's not very good, if anyone has a better idea please tell !

src/reachy_mini_conversation_app/main.py
apirrone: I don't think it's an issue since `wrapped_run()` created a `ReachyMini` Instance everytime you run an app 

https://github.com/pollen-robotics/reachy_mini/blob/d51a24734bce0db2184dc3b11f286a441938a6b8/src/reachy_mini/apps/app.py#L84

src/reachy_mini_conversation_app/main.py
alozowski: Makes sense then!

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: Let's try with negative args first? At least it's safer, a bit

src/reachy_mini_conversation_app/main.py
alozowski: Got it!

src/reachy_mini_conversation_app/main.py
apirrone: Done

PR #123, opened 2025-12-01
fix(camera tool): removing temporary file creation in the Camera tool

CarolinePascal:
The PR removed the temporary file creation in the Camera tool, which was not compatible with Windows. The current solution encodes the frame in the JPG format but stays in the RAM.


CarolinePascal:
Successfully tested on Windows + MacOS

PR #131, opened 2025-12-03
Update stop function to reset to neutral position

alozowski:
## Summary
When stopping the conversation app, the robot now smoothly returns to a neutral position (head centered, antennas at 0 degree) over 1.5 seconds before shutting down, preventing it from ending in awkward or extreme poses. This is implemented by queuing a `GotoQueueMove` in the `MovementManager.stop()` method that interpolates from the current position to neutral before terminating the control loop.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [x] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
- [x] Everything is tested in simulation as well (`--gradio` required)

### Vision / motion
- [ ] Head wobble

### Dependencies & config
- [ ] Updated `pyproject.toml` if deps/extras changed
- [ ] Regenerated `uv.lock` if deps changed

### DIFF
--- a/src/reachy_mini_conversation_app/moves.py
+++ b/src/reachy_mini_conversation_app/moves.py
@@ -48,7 +48,6 @@ from reachy_mini.utils import create_head_pose
 from reachy_mini.motion.move import Move
 from reachy_mini.utils.interpolation import (
     compose_world_offset,
-    distance_between_poses,
     linear_pose_interpolation,
 )
 


src/reachy_mini_conversation_app/moves.py
```diff
@@ -723,7 +723,62 @@ def start(self) -> None:
...
+        # Clear any queued moves and stop current move
+        self.clear_move_queue()
+
+        # Create a simple move to neutral position
+        try:
+            from reachy_mini_conversation_app.dance_emotion_moves import GotoQueueMove
+
+            # Get current position to interpolate from
+            try:
+                _, current_antennas = self.current_robot.get_current_joint_positions()
```
RemiFabre: Note for another time: this whole block of code is the correct thing to do but it's very verbose and it's used in several places. We should create a "create_goto_move" function for this

src/reachy_mini_conversation_app/moves.py
```diff
@@ -723,7 +723,62 @@ def start(self) -> None:
...
+
+            # Create a goto move to neutral with 1.5 second duration
+            reset_move = GotoQueueMove(
+                target_head_pose=neutral_head_pose,
+                start_head_pose=current_head_pose,
+                target_antennas=neutral_antennas,
+                start_antennas=current_antennas,
+                target_body_yaw=neutral_body_yaw,
+                start_body_yaw=0.0,
+                duration=1.5,
```
RemiFabre: Using a fixed duration creates a few problems. In this case, what would happen if the robot was already in the neutral position? --> We would wait 1.8 seconds being immobile.
And if the robot was close to the final pose, then it would move suuuper slowly.

Instead, we should scale the duration based on the "distance". I'm writing "distance" here because... ideally its not just a l2 distance (with xyz) but also integrates the angular distance (rpy). I like to call this the "magic_distance" for fun :) Imo you should copy this:
https://github.com/pollen-robotics/reachy_mini/blob/04b37a6255918d83d2c0e78e4fc7c8172248a1b5/src/reachy_mini/daemon/backend/abstract.py#L644

RemiFabre (changes requested):

src/reachy_mini_conversation_app/moves.py
alozowski: I'll leave it as a TODO comment

src/reachy_mini_conversation_app/moves.py
alozowski: You're right, I didn't think about all the edge cases – addressed this 👍🏻

src/reachy_mini_conversation_app/moves.py
```diff
@@ -723,7 +724,78 @@ def start(self) -> None:
...
+            neutral_body_yaw = 0.0
+
+            # Calculate the magic distance and scale duration accordingly
+            _, _, magic_distance = distance_between_poses(current_head_pose, neutral_head_pose)
+
+            # Duration scaling: 20 ms per magic_mm (same as wake_up)
+            # Add a minimum duration to avoid instant jumps, and a maximum to avoid waiting forever
+            ms_per_magic_mm = 20
+            duration = magic_distance * ms_per_magic_mm / 1000  # convert ms to seconds
+            duration = max(0.2, min(duration, 2.0))  # clamp between 0.2s and 2.0s
```
RemiFabre: I can agree with the clamping at 2.0 but I find the 0.2 minimal delay weird. Imagine that the robot is already at the right position, why wait 0.2s doing nothing?
Also the implementation of the "wake_up" function has be tested a loooot of times, so I'd replicate it with confidence.
Imo just replace 0.2 with 0.0 and we're good. Sorry for being annoying with this.

alozowski: No worries, thanks a lot for your comments! Got the idea 👍🏻

PR #143, opened 2025-12-15
improve the startup with openai

andimarafioti:
## Summary
This app improves the startup of the app by automatically providing users with an openai key if they don't have one. 

## Category
- [x] Fix
- [x] Feature

### Run modes
- [x] Headless mode (default)
- [x] Gradio UI (`--gradio`)
- [x] Everything is tested in simulation as well (`--gradio` required)

### Dependencies & config
- [x] Updated `pyproject.toml` if deps/extras changed
- [x] Regenerated `uv.lock` if deps changed

## Notes
We might decide to undo this change eventually, but for now it's a good compromise


alozowski (approved):
it's cool!

PR #163, opened 2025-12-17
Improve interruptions

andimarafioti:
## Summary
Fix interruptions. Because OpenAI sends chunks of audio, I saw in the logs that we were interrupting the robot, and everything seemed to work, except that the robot would continue to speak. I suspected this was the output buffer from audiosounddevice not beeing emptied, so I added that. This PR requires [this one in reachy mini](https://github.com/pollen-robotics/reachy_mini/pull/574) to be merged.

## Category
- [X] Fix


CarolinePascal (approved):
Nice, let's merge too !

PR #192, opened 2026-01-22
Enable loading custom profiles and tools from outside the library

A-Mahla:
## Summary

This PR enables users to define custom profiles and tools **outside** the `reachy_mini_conversation_app` library, making it easier to extend the app without modifying the source code.
### Changes

External profiles and tools

You can extend the app with profiles/tools stored outside `src/reachy_mini_conversation_app/`.

- Core profiles are under `src/reachy_mini_conversation_app/profiles/`.
- Core tools are under `src/reachy_mini_conversation_app/tools/`.

**Recommended layout:**

```text
external_content/
├── external_profiles/
│   └── my_profile/
│       ├── instructions.txt
│       ├── tools.txt        # optional (see fallback behavior below)
│       └── voice.txt        # optional
└── external_tools/
    └── my_custom_tool.py
```

**Environment variables:**

Set these values in your `.env` (copy from `.env.example`):

```env
REACHY_MINI_CUSTOM_PROFILE=my_profile
REACHY_MINI_EXTERNAL_PROFILES_DIRECTORY=./external_content/external_profiles # or custom path
REACHY_MINI_EXTERNAL_TOOLS_DIRECTORY=./external_content/external_tools # or custom path
# Optional convenience mode:
# AUTOLOAD_EXTERNAL_TOOLS=1
```

**Loading behavior:**

- **Default/strict mode**: `tools.txt` defines enabled tools explicitly. Every name in `tools.txt` must resolve to either a built-in tool (`src/reachy_mini_conversation_app/tools/`) or an external tool module in `REACHY_MINI_EXTERNAL_TOOLS_DIRECTORY`.
- **Convenience mode** (`AUTOLOAD_EXTERNAL_TOOLS=1`): all valid `*.py` tool files in `REACHY_MINI_EXTERNAL_TOOLS_DIRECTORY` are auto-added.
- **External profile fallback**: if the selected external profile has no `tools.txt`, the app falls back to built-in `profiles/default/tools.txt`.
### Backward Compatibility

Fully backward compatible — existing profiles and tools within the library continue to work without any configuration changes.

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [x] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config
- [x] Updated `.env.example` if new config vars added


A-Mahla:
Nice point @alozowski! I really like the “tools store on the Hub” idea. It could make the conversational app a truly modular building block for Reachy Mini builders, especially for non-technical users (install/uninstall, discovery, etc.).

That said, I think this is a bit orthogonal to the scope of this PR (sorry if my context wasn’t clear). The goal here is simply to **reduce friction for developers**: allow custom tools/profiles to be defined outside the library, using only `import reachy_mini_conversational_app`, and run the conversational app main script (with defined new env variables) to load their custom tools. This way, devs can iterate and integrate tools on their side without forking or editing the core repo. This PR is intentionally “dev-first” and doesn’t try to solve distribution/UX for non-dev users yet.

My proposal would be:
	1.	Merge this PR to unlock external tool creation for developers (minimal surface area, no breaking changes).
	2.	Then follow up with a dedicated PR to design a robust Hub-based install/uninstall flow (and the safety layer), which deserves its own discussion/spec.

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     #Reachy mini
     "reachy_mini_dances_library",
     "reachy_mini_toolbox",
-    "reachy-mini >= 1.2.11",
+    "reachy-mini @ git+https://github.com/pollen-robotics/reachy_mini.git@develop",
```
A-Mahla: Are we sure we want to change the module source in the  `reachy_mini_conversation_app` develop branch ?

CarolinePascal: The is indeed likely to break the app installation because of `uv` and `git lfs`! You must revert this change.

PR #195, opened 2026-01-23
Adding a LOCKED_PROFILE constant in config.py to help forking the app into a specialized app.

scienceetonnante:
## Summary
Adds a LOCKED_PROFILE constant in config.py. By default the constant is set to None, this preserves the normal behavior : profile is managed by the env variable, can be set in the gradio UI, etc.

If LOCKED_PROFILE is set to a profile, this will lock the app on this profile and disable the gradio UI to modify it.

**Purpose :** When forking the conversation app to make a new specialized app with a dedicated profile, this constant can be set by the dev (or the cloning script) in order to lock the profile of the new app.

## Category
- [ ] Fix
- [X] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic
- [X] CI green (Ruff, Tests, Mypy)
- [X] Code update is clear (types, docs, comments)

### Run modes
- [X] Headless mode (default)
- [X] Gradio UI (`--gradio`)
- [X] Everything is tested in simulation as well (`--gradio` required)

### Vision / motion
- [ ] Head wobble

### Dependencies & config

### DIFF
--- a/src/reachy_mini_conversation_app/config.py
+++ b/src/reachy_mini_conversation_app/config.py
@@ -1,5 +1,7 @@
 import os
+import sys
 import logging
+from pathlib import Path
 
 from dotenv import find_dotenv, load_dotenv
 
@@ -9,6 +11,18 @@ from dotenv import find_dotenv, load_dotenv
 LOCKED_PROFILE: str | None = None
 
 logger = logging.getLogger(__name__)
+
+# Validate LOCKED_PROFILE at startup
+if LOCKED_PROFILE is not None:
+    _profiles_dir = Path(__file__).parent / "profiles"
+    _profile_path = _profiles_dir / LOCKED_PROFILE
+    _instructions_file = _profile_path / "instructions.txt"
+    if not _profile_path.is_dir():
+        print(f"Error: LOCKED_PROFILE '{LOCKED_PROFILE}' does not exist in {_profiles_dir}", file=sys.stderr)
+        sys.exit(1)
+    if not _instructions_file.is_file():
+        print(f"Error: LOCKED_PROFILE '{LOCKED_PROFILE}' has no instructions.txt", file=sys.stderr)
+        sys.exit(1)
 
 # Locate .env file (search upward from current working directory)
 dotenv_path = find_dotenv(usecwd=True)


src/reachy_mini_conversation_app/gradio_personality.py
```diff
@@ -85,23 +85,33 @@ def _sanitize_name(name: str) -> str:
...
-        self.new_personality_btn = gr.Button("New personality")
-        self.available_tools_cg = gr.CheckboxGroup(label="Available tools (helper)", choices=[], value=[])
-        self.save_btn = gr.Button("Save personality (instructions + tools)")
+        self.person_name_tb = gr.Textbox(label="Personality name", interactive=not is_locked)
+        self.person_instr_ta = gr.TextArea(label="Personality instructions", lines=10, interactive=not is_locked)
+        self.tools_txt_ta = gr.TextArea(label="tools.txt", lines=10, interactive=not is_locked)
+        self.voice_dropdown = gr.Dropdown(label="Voice", choices=["cedar"], value="cedar", interactive=not is_locked)
+        self.new_personality_btn = gr.Button("New personality", interactive=not is_locked)
+        self.available_tools_cg = gr.CheckboxGroup(label="Available tools (helper)", choices=[], value=[], interactive=not is_locked)
+        self.save_btn = gr.Button("Save personality (instructions + tools)", interactive=not is_locked)
```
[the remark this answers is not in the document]
alozowski: I agree with copilot 🤖

src/reachy_mini_conversation_app/config.py
```diff
@@ -4,6 +4,10 @@
 from dotenv import find_dotenv, load_dotenv
 
 
+# Locked profile: set to a profile name (e.g., "astronomer") to lock the app
+# to that profile and disable all profile switching. Leave as None for normal behavior.
+LOCKED_PROFILE: str | None = None
```
alozowski, line 11: If `LOCKED_PROFILE` is set to a profile that doesn’t exist, `get_session_instructions()` exits the process during startup. That's a late, runtime failure instead of a clear config error. I suggest to validate the locked profile directly here in `config.py` so the app fails fast with a predictable message rather than crashing deeper in the session setup

src/reachy_mini_conversation_app/gradio_personality.py
```diff
@@ -85,23 +85,33 @@ def _sanitize_name(name: str) -> str:
...
+            dropdown_label = "Select personality"
+            dropdown_choices = [self.DEFAULT_OPTION, *(self._list_personalities())]
 
         self.personalities_dropdown = gr.Dropdown(
-            label="Select personality",
-            choices=[self.DEFAULT_OPTION, *(self._list_personalities())],
+            label=dropdown_label,
+            choices=dropdown_choices,
             value=current_value,
+            interactive=not is_locked,
```
alozowski, line 103: Here Gradio UI disables the "Apply personality" button, but `_apply_personality` itself doesn't check for the lock, so it could still be triggered programmatically. Would you consider adding a small guard inside `_apply_personality` for defense-in-depth (similar to the headless `/personalities/apply` endpoint)? It's not critical, but it keeps behavior consistent across interfaces

RemiFabre (approved):

Issue #201, opened 2026-02-06
Camera tool logic is double and suboptimal


scienceetonnante:
The camera tool does two very different things depending on whether local vision is enabled or not. The tool has one mandatory argument `question`

* with local vision, the tool calls the local vision model with the question, the local model returns a string description that is then being returned to the LLM as a tool call results, the LLM can then follow up (and this is the expected way to handle tool call).

* without local vision, the `question` is ignored, the tool call result first returns the b64 encoding of the image (that the OpenAI realtime API can't use apparently !, but it costs token usage and context space), then send again the image in a usable format as a user message. So the image is sent twice, but the first time, the tool call results is unsable and confusing for the LLM.

I suggest creating two distinct tools with clear contract for the LLM:

* One to use with local vision model, that clearly exposes that the output will be a string answering a question
* One to use without vision, that does not take a `question` argument, does not return unusable b64, but makes it clear that user will send the image right after.

We can keep camera for backward compatibility.

alozowski:
Closing as this has been resolved in two steps: the local vision path was removed entirely in #430, so the camera tool now has a single behaviour, and since #279 the b64 is stripped from the tool result before it's echoed to the model (replaced with `image_attached: true`), with the image attached exactly once as a proper `input_image` user message. The tool still takes a question argument that's now only a focus hint

PR #202, opened 2026-02-06
Introduces two distinct tools to be used instead of "camera"

scienceetonnante:
Answers the following issue #201 

I introduced two new tools to be used instead of the "camera" tool, which had a dual behavior with suboptimal logic (wasted tokens and cluttered context) when used without local vision.

## Summary

With those two new tools, the two behavior are clearly separated, with their own logic and their own data flow:

`describe_camera_image` : this tool takes a "question" as an argument, call a vision model and returns to the LLM the string description. This is essentially the same behavior as the camera tool when a local vision model is present.

`see_image_through_camera`: does not take any argument, and feed the captured image to the LLM conversation.

camera tool is still here for retrocompatibility.

For now only the default tools.txt is modified to comment out "camera" and add those two new tools.

## Category
- [X] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic
- [X] CI green (Ruff, Tests, Mypy)
- [X] Code update is clear (types, docs, comments)

### Run modes
- [X] Headless mode (default)
- [X] Gradio UI (`--gradio`)

### Vision / motion
- [X] Local vision (`--local-vision`)
- [X] Camera pipeline (with/without `--no-camera`)
- [ ] Head wobble
- [X] Profiles or custom tools

### Dependencies & config

## Notes

For now the exclusion list is such that either one or the other tool is exposed. But we could imagine that when a local vision is present AND the LLM is multimodal, we expose both tools and leave it to the LLM to decide whether to call one or the other. But I preferred to avoid this for now.


andimarafioti:
Maybe also remove the old camera tool if you want to replace it?

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -257,7 +257,10 @@ async def _run_realtime_session(self) -> None:
                                 "voice": get_session_voice(),
                             },
                         },
-                        "tools": get_tool_specs(),  # type: ignore[typeddict-item]
+                        "tools": get_tool_specs(exclusion_list=(
```
alozowski: This works, but now the session setup code has to "know" which tool needs which dependency and if another part of the codebase calls `get_tool_specs`, it would need to copy-paste this same logic.
I can suggest two ideas:
- add a small helper like `_camera_tool_exclusions(deps) -> list[str]` to keep it in one place
- add a `requires_vision:` bool flag on the Tool class so `get_tool_specs` can figure it out on its own

I prefer the second option though it requires more code changes, but it will be cleaner

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -376,37 +379,44 @@ async def _run_realtime_session(self) -> None:
...
                         logger.debug("Tool '%s' executed successfully", tool_name)
-                        logger.debug("Tool result: %s", tool_result)
+                        logger.debug("Tool result: %.500s", tool_result)
                     except Exception as e:
                         logger.error("Tool '%s' failed", tool_name)
                         tool_result = {"error": str(e)}
 
+                    # For see_image_through_camera/camera with b64 result: send a lightweight tool result
+                    # and inject the image as a proper user message instead.
+                    is_image_tool = tool_name in ("camera", "see_image_through_camera") and "b64_im" in tool_result
```
alozowski: I agree with @andimarafioti's question about removing the old camera tool and if you decide to remove it, it's possible to simplify this line as:
```
is_image_tool = isinstance(tool_result, dict) and "b64_im" in tool_result
```

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -376,37 +379,44 @@ async def _run_realtime_session(self) -> None:
...
+                    if is_image_tool:
+                        b64_im = tool_result["b64_im"]
+                        if not isinstance(b64_im, str):
+                            logger.warning("Unexpected type for b64_im: %s", type(b64_im))
+                            b64_im = str(b64_im)
```
alozowski, line 393: I would add `# see_image_through_camera guarantees str, but guard against future tools` comment here if we keep only `see_image_through_camera` tool

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -376,37 +379,44 @@ async def _run_realtime_session(self) -> None:
...
+                        tool_result_for_api = {"status": "Image captured, will be provided next."}
```
alozowski: This message gets sent to the LLM, so its wording could matter for behavior. Consider pulling it into a module-level constant, for example, as a IMAGE_PLACEHOLDER_RESULT, so it's easy to find and tweak later.

src/reachy_mini_conversation_app/tools/see_image_through_camera.py
```diff
@@ -0,0 +1,43 @@
...
+            return {"error": "Camera worker not available"}
+
+        frame = deps.camera_worker.get_latest_frame()
+        if frame is None:
+            logger.error("No frame available from camera worker")
+            return {"error": "No frame available"}
+
+        success, buffer = cv2.imencode(".jpg", frame)
+        if not success:
+            raise RuntimeError("Failed to encode frame as JPEG")
```
alozowski: Every other failure in the tools returns `{"error": ...}`, but this one raises an exception. The caller does catch it, so it won't crash, but it's inconsistent and could confuse someone adding a new caller later
p.s. I see it was actually inherited from the old `camera.py` which did the same thing

src/reachy_mini_conversation_app/tools/see_image_through_camera.py
```diff
@@ -0,0 +1,43 @@
...
+
+class SeeImageThroughCamera(Tool):
+    """Take a picture with the camera so you can see what's in front of you."""
+
+    name = "see_image_through_camera"
+    description = "Get a picture taken with the camera in your eyes. Use this so you can see what's in front of you."
+    parameters_schema = {
+        "type": "object",
+        "properties": {},
+    }
```
alozowski, line 24: `describe_camera_image` has `"required": ["question"]`. For consistency (and because some schema validators care), it'd be nice to add `"required": []` here too

src/reachy_mini_conversation_app/tools/describe_camera_image.py
```diff
@@ -0,0 +1,59 @@
...
+
+        vision_result = await asyncio.to_thread(
+            deps.vision_manager.processor.process_image, frame, image_query,
+        )
+        if isinstance(vision_result, dict) and "error" in vision_result:
+            return vision_result
+        if isinstance(vision_result, str):
+            logger.info("Vision model response: %s", vision_result[:500])
+            return {"description": vision_result}
+        return {"error": "vision returned non-string"}
```
alozowski: If `process_image` returns something unexpected, "vision returned non-string" doesn't tell you much when debugging, so we can include the type like this:
```
return {"error": f"vision returned unexpected type: {type(vision_result).__name__}"}
```

(I didn't test this approach myself, please, test it)

src/reachy_mini_conversation_app/tools/describe_camera_image.py
```diff
@@ -0,0 +1,59 @@
...
+            return {"error": "No frame available"}
+
+        vision_result = await asyncio.to_thread(
+            deps.vision_manager.processor.process_image, frame, image_query,
+        )
+        if isinstance(vision_result, dict) and "error" in vision_result:
+            return vision_result
+        if isinstance(vision_result, str):
+            logger.info("Vision model response: %s", vision_result[:500])
+            return {"description": vision_result}
```
alozowski, line 59: The old `camera.py` returned `{"image_description": vision_result}`, this returns `{"description": vision_result}`. That's fine because nothing external reads this key but worth being aware of

alozowski (changes requested):
I left some comments and if we assume we remove the old `camera.py`, there are 13 profiles still reference camera:
```
❯ grep -r "camera" src/reachy_mini_conversation_app/profiles/*/tools.txt
src/reachy_mini_conversation_app/profiles/cosmic_kitchen/tools.txt:camera
src/reachy_mini_conversation_app/profiles/default/tools.txt:#camera
src/reachy_mini_conversation_app/profiles/default/tools.txt:see_image_through_camera
src/reachy_mini_conversation_app/profiles/default/tools.txt:describe_camera_image
src/reachy_mini_conversation_app/profiles/example/tools.txt:# camera
src/reachy_mini_conversation_app/profiles/mars_rover/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_bored_teenager/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_captain_circuit/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_chess_coach/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_hype_bot/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_mad_scientist_assistant/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_nature_documentarian/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_noir_detective/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_time_traveler/tools.txt:camera
src/reachy_mini_conversation_app/profiles/short_victorian_butler/tools.txt:camera
src/reachy_mini_conversation_app/profiles/sorry_bro/tools.txt:camera
```

Thus, we need to be careful and replace `camera` with `see_image_through_camera` and `describe_camera_image` in each `.txt`

Let's also add some tests to cover the new changes or camera capabilities in general.

scienceetonnante:
Ok so I removed the camera tool and adapted the tools.txt list for each profile.
I added a test and also fixed the minor stuff.

alozowski:
Closing this along with #201: the problem it addressed got solved by other changes lately. The b64 waste was fixed in #279 (the tool result now carries `image_attached: true` and the image is attached once as a user message), and the dual local-vision path disappeared entirely when local vision was removed in #430. The files this PR touches no longer exist after #444, so a rebase isn't meaningful

PR #204, opened 2026-02-11
Add serial response.create worker to prevent Realtime API conflicts

A-Mahla:
## Summary

- Introduces a dedicated `_response_sender_loop` worker that serializes all `response.create()` calls to the OpenAI Realtime API, ensuring only one response is active at a time
- Replaces direct `connection.response.create()` calls with a `_safe_response_create()` queue-based method that never blocks the caller
- Automatically retries rejected requests when the server returns `conversation_already_has_active_response`, instead of treating it as a hard error

### Problem

The OpenAI Realtime API only allows one active response per conversation at a time. When multiple parts of the code (e.g., tool-call follow-ups and idle behavior triggers) race to call `response.create()`, the server rejects the second call with a `conversation_already_has_active_response` error. This caused dropped responses and noisy error logs.

### Solution

A new async worker task (`_response_sender_loop`) consumes from an `asyncio.Queue` and sends `response.create()` requests one at a time. It:

1. Waits for the previous response cycle to complete (`_response_done_event`)
2. Sends the next queued request
3. If the server rejects it, waits for the active response to finish and retries
4. Properly shuts down when the connection closes or `shutdown()` is called

All call-sites now go through `_safe_response_create()` instead of calling the connection directly.

## Category
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config


alozowski (changes requested):
Reviewed here:
https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/205#pullrequestreview-3850138363

PR #205, opened 2026-02-12
Concurrent Tool Execution - add `BackgroundToolManager`

A-Mahla:
## Context

Tools were blocking, so any tool call froze the event loop,the robot couldn't talk, move, or respond. After ~20s the OpenAI websocket timed out and the connection dropped.

This PR solves the problem by running all tool calls as background async tasks, keeping the event loop free to maintain the websocket connection and continue the conversation while tools execute.

## Summary

- **Introduce `BackgroundToolManager`** to run tool calls asynchronously in the background, allowing the robot to continue conversing while long-running tools (e.g., camera, weather) execute without blocking the realtime audio session.
- **Add system tools `task_status` and `task_cancel`** so the LLM can introspect and manage running background tasks (check progress, cancel stalled tools).

## Key Changes

| File | Change |
|---|---|
| `background_tool_manager.py` (new) | Core orchestrator: `BackgroundToolManager` with async task spawning, notification queue, automatic timeout/cleanup, and progress tracking |
| `tool_constants.py` (new) | `ToolState` and `SystemTool` enums for status tracking and system tool auto-registration |
| `task_status.py` / `task_cancel.py` (new) | System tools that let the LLM query running tasks and cancel them by ID |
| `openai_realtime.py` | Refactored tool dispatch: tools are now launched via `BackgroundToolManager.start_tool()`. Results are handled asynchronously through `_handle_tool_result` callback. The robot immediately notifies the user a tool is running rather than going silent. |
| `core_tools.py` | `dispatch_tool_call` now accepts `tool_manager` arg for system tools. |
| `default_prompt.txt` | Added instruction to prevent duplicate tool calls execution with identical arguments |
| `test_background_tool_manager.py` (new) | Full async test suite covering models, lifecycle, progress, cancellation, timeout, cleanup, getters, startup, and notification queue |

## How It Works

1. When a `response.function_call_arguments.done` event arrives, the tool is registered and launched as an `asyncio.Task` via `BackgroundToolManager.start_tool()`.
2. The robot immediately tells the user the tool is running (non-blocking UX).
3. When the tool completes/fails, a `ToolNotification` is pushed to an internal queue.
4. A background listener picks up the notification and calls `_handle_tool_result`, which sends the result back to the OpenAI conversation and triggers a spoken response.
5. Periodic cleanup auto-cancels tools running past `max_tool_duration_seconds` (default: 1h) and purges finished tool records past `max_tool_memory_seconds` (default: 1 day).

## Demo

https://github.com/user-attachments/assets/956cf4b0-6eb3-48ec-ae38-1d5614194250

You can try it with the custom `GetWeather` tool from the demo in the branch: `amir/test_concurrent_execution_tool`. 

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [x] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config


alozowski (changes requested):
Hi @A-Mahla!

I see that this PR introduces non-blocking tool execution for all tools _but_ we might want some tools to remain blocking – specifically, we need to discuss about all the robot's movement tools. We might also want certain tools to block each other, for example, we don't want the camera to take a picture while the robot is in motion. 

Right now PR [#204](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/204) fixes a race condition that this PR 205 introduces, so we need to restructure.

The first step is to work on _a solid base PR that implements non-blocking tool execution correctly_ – including the response serialisation, so that everything after it only adds functionality on top, without needing to fix the base. Then break the rest into smaller PRs, _all targeting develop_:
- non-blocking tool execution + response serialisation – `BackgroundToolManager`, `tool_constants.py`, `_response_sender_loop`,` _safe_response_create`, integration into `openai_realtime.py`, and tests for all of it. This is the foundation – it needs to work on its own.
- system tools – `task_status.py`, `task_cancel.py`, `core_tools.py` changes. Depends on p.1
- any further features, for example prompt changes – depends on above.

I also found some _unrelated changes_ that should be reverted. Please, _do not_ modify working code that has nothing to do with non-blocking execution:
- `core_tools.py` – `dispatch_tool_call` error handling: changing `return {"error": msg}` to `raise e` breaks the existing contract. Every caller expects an error dict back, not an exception.
- `core_tools.py` – `dispatch_tool_call` signature: the `**kwargs` addition to smuggle `tool_manager` through the args dict changes the public interface and mixes infrastructure with tool arguments. Please, find a cleaner way to pass the manager for system tools.
- `core_tools.py` – `list` to `set` in `_load_profile_tools`: unrelated refactor that also changes logging behavior (reports all tools instead of only newly discovered ones)
- `test_openai_realtime.py` – removed `asyncio.sleep` monkeypatch: unrelated to the feature
- `default_prompt.txt`: prompt behavior change, not related to the execution mechanism. Please, move to a separate PR.

Issues I found in the new code:
- `openai_realtime.py`: `conversation_already_has_active_response` was removed from the UI error exclusion set – please, add it back, users shouldn't see these.
- Double `response.create()` per tool call: old code sent one response after the tool result, new code sends two (one "tool is running", one with the result). If intentional, please, explain it. If not, remove the first one.
- `_handle_tool_result`: stray comma on its own line (`"status": "done"\n,`). Also, the `ConnectionClosedError` catch should call `self._response_done_event.set()` to avoid deadlocking the sender loop.
- `_response_sender_loop`: no retry cap on the inner loop – add a limit (3–5) with backoff.
- `background_tool_manager.py`: `set_loop` logs at WARNING – should be INFO or DEBUG.
- Type annotation: declare `self.connection: AsyncRealtimeConnection | None = None` in `__init__` instead of `# type: ignore[no-redef]`

Please close [PR 204](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/204) and let's rework this PR 205 first to make it a base one.

src/reachy_mini_conversation_app/tools/background_tool_manager.py
```diff
@@ -0,0 +1,412 @@
...
+            bg_tool.completed_at = time.monotonic()
+
+            logger.info(f"Background tool completed: {bg_tool.tool_name} (id={bg_tool.id})")
+
+        except asyncio.CancelledError:
+            bg_tool.status = ToolState.CANCELLED
+            bg_tool.completed_at = time.monotonic()
+            bg_tool.error = "Tool cancelled"
+            logger.info(f"Background tool cancelled: {bg_tool.tool_name} (id={bg_tool.id})")
+            raise
```
RemiFabre: Cancelled tools never send `function_call_output` is that normal? Maybe we should put the call in a finally block

src/reachy_mini_conversation_app/tools/background_tool_manager.py
```diff
@@ -0,0 +1,412 @@
...
+                for callback in tool_callbacks:
+                    await callback(bg_tool)
+
+        async def _cleanup(interval_seconds: float = 5 * 60) -> None:
+            while True:
+                await asyncio.sleep(interval_seconds)
+                await self.cleanup_tools()
+                await self.timeout_tools()
+
+        self._lifecycle_tasks = [
```
RemiFabre, line 313: I'm not 100% sure but could you please review this potential problem:
Every time we do a _restart_session(), these tasks are created again (the old ones are not pointed in this list anymore but might still be alive), and having 2 listeners could create some nasty bugs.
If shutdown() is called before start_up() then it's fine, but I'm not sure if it always the case?

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -403,92 +521,52 @@ async def _run_realtime_session(self) -> None:
...
-                                    },
-                                ),
-                            )
-
-                    # if this tool call was triggered by an idle signal, don't make the robot speak
-                    # for other tool calls, let the robot reply out loud
                     if self.is_idle_tool_call:
                         self.is_idle_tool_call = False
                     else:
                         await self.connection.response.create(
```
RemiFabre: Weirdly this doesn't seem coherent to what I experienced when testing? But my understanding of the code (CTRL+F "self.connection.response.create") is that the LLM is asked to comment on tool start and tool end. That's gong to get old super fast for short emotions/dances/etc.

src/reachy_mini_conversation_app/tools/background_tool_manager.py
```diff
@@ -0,0 +1,412 @@
...
+    """the completion time of the tool"""
+    completed_at: Optional[float] = None
+
+    """the async tool execution task"""
+    _task: Optional[asyncio.Task[None]] = PrivateAttr(default=None)
+
+    @property
+    def tool_id(self) -> str:
+        """Get the name of the tool."""
+        return f"{self.tool_name}-{self.id}"
```
RemiFabre: Are we sure each `call_id` is unique? If not, it could lead to some ugly bugs. Maybe add a safeguard just in case?

RemiFabre (changes requested):

src/reachy_mini_conversation_app/tools/background_tool_manager.py
A-Mahla: fixed

src/reachy_mini_conversation_app/tools/background_tool_manager.py
A-Mahla: I shut down these tasks if the connection closes in[ _run_realtime_session()](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/cbc0c4d1079d8327f152e0a84566a2364c4e850e/src/reachy_mini_conversation_app/openai_realtime.py#L678)

src/reachy_mini_conversation_app/tools/background_tool_manager.py
A-Mahla: added the [`BackgroundTool.started_at` timestamp to the `tool_id` ](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/023b8a5aca2cf8edfdb0477fcad3fab4ec460994/src/reachy_mini_conversation_app/tools/background_tool_manager.py#L100)

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Yes, that’s the intended behavior. The logic already existed when you've test and I’ve also tested it. For some reason (LLM side), I didn’t see fast tools (emotions, dances, etc.) announce that "they’d started" but they indeed speak about their action.

I can remove it, but here’s the rationale: this is still an open behavior question. For long-running tasks, it’s perfect because it notifies the user via speech that the task has started. For short-running tasks, it can indeed become a bit verbose.

If you have a better UX idea to make sure the user knows when a tool starts (e.g., announce only for long-running tools and stay silent for very fast ones), I’m very open to suggestions. BTW, you can see in the video how Reachy notifies the user when the GetWeather mock tool is called.

RemiFabre: Yes. I think let's keep it that way since it was already like that. I've changed this in several experimental branches. Maybe we'll eventually change this, but this is out of the scope of this PR.

RemiFabre (approved):

Issue #214, opened 2026-02-20
Integrate tool store


alozowski:
Add tool store support so users can browse and download external tools. This should be accessible both from the unified UI (see https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/213) and via CLI.

PR #222, opened 2026-02-26
Add external tools download from Hugging Face Spaces

alozowski:
## Summary
This PR adds a simple way to import external tools from a HF Space into the app. Related to https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/214. 

With `--download-hf-tool owner/repo`, the app downloads one tool module, syncs safe dependencies if `requirements.txt` exists, writes tool metadata, and exits. I tried to ensure a safe flow as well.

- added new CLI flag: `--download-hf-tool <owner/repo>`.
- added sync workflow in `tool_dependency_sync.py` for a Space ID normalisation, revision pinning, tool download, and metadata creation.
- added safety checks for `requirements.txt` entries (no unsafe directives/URLs/local paths).
- added rollback logic for `pyproject.toml` and `uv.lock` if sync fails.
- added duplicate `Tool.name` detection when loading tools.
- updated README and tests for the new CLI flow and failure/success cases.

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config


src/reachy_mini_conversation_app/tool_dependency_sync.py
```diff
@@ -0,0 +1,600 @@
...
+_UNSAFE_REQUIREMENTS_SUBSTRINGS = (
+    "://",
+    "git+",
+    "svn+",
+    "hg+",
+    "bzr+",
+    "file:",
+    " @ ",
+)
+_UNSAFE_REQUIREMENTS_PATH_PREFIXES = (
```
alozowski, line 44: I'm pretty sure that this won't work on Windows – I need to think

alozowski: solved

src/reachy_mini_conversation_app/tool_dependency_sync.py
```diff
@@ -0,0 +1,604 @@
...
+            raise ValueError(
+                f"Unsafe requirements.txt entry at line {line_no}: '{requirement}'. "
+                "Local file paths are not allowed for external tool dependency sync."
+            )
+
+        if any(token in lowered for token in _UNSAFE_REQUIREMENTS_SUBSTRINGS):
+            raise ValueError(
+                f"Unsafe requirements.txt entry at line {line_no}: '{requirement}'. "
+                "Direct URL/VCS/file references are not allowed for external tool dependency sync."
+            )
```
A-Mahla, line 140: Just a thought: I could publish a malicious package to PyPI and import it in my tool. Do we have, or are we planning to have, any sanitization or restriction strategy for imports? Small risk, but it exists. I think there are some security tools for this on HF Hub side.

alozowski: yes, I though about a malicious pypi package, I can check if there are existing approaches

RemiFabre:
Hello,
As it stands, this adds the ability to download a tool locally and then exists. I would make it a separate script since it has nothing to do with the execution of the conversation app. 

This is a cool feature for devs, let's see if it catches on! If it does, maybe it will be worth it to consider a more ambitious take where it's possible to dynamically download tools, even through the web GUI for non experts (but I won't explore this right now).

alozowski:
Yes, I also don't like that the current implementation requires to pass `reachy-mini-conversation-app ...` only for a downloading process that has nothing to do with the conv app itself – I'll try to separate it into a something standalone but inside the project

alozowski:
I came up with a better idea! 😈

PR #227, opened 2026-03-03
Update default prompt for background tool execution

A-Mahla:
## Summary

References https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/205. This is part 3 of 3, prompt to avoid duplicate consecutive tool calls unless explicitly requested by the user.

## Category
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config

### DIFF
--- a/src/reachy_mini_conversation_app/prompts/default_prompt.txt
+++ b/src/reachy_mini_conversation_app/prompts/default_prompt.txt
@@ -37,6 +37,7 @@ Keep safety in mind when giving guidance.
 
 ## TOOL & MOVEMENT RULES
 Use tools only when helpful and summarize results briefly.
+Never call the same tool with identical arguments twice in a row. Always wait for the result before calling again. Duplicate calls waste resources and cause errors. The only exception is when the user explicitly requests it.
 Use the camera for real visuals only — never invent details.
 The head can move (left/right/up/down/front).
 


src/reachy_mini_conversation_app/prompts/default_prompt.txt
```diff
@@ -37,6 +37,7 @@ Keep safety in mind when giving guidance.
 
 ## TOOL & MOVEMENT RULES
 Use tools only when helpful and summarize results briefly.
+Never call the same tool with identical arguments twice in a row. Always wait for the result before calling again. Duplicate calls waste resources and cause errors. The only exception is when the user explicitly requests it.
```
A-Mahla, line 40: Changed the prompt because, with background execution, I noticed a behavior where the model sometimes sends the same tool call multiple times in a row with the same arguments before receiving the response.

RemiFabre: wow, this is weird. What's your explanation for this behavior? Maybe unrelated but this makes me think of the potential bug I wrote about on your other PR where 2 producers could be alive in //. Could this me it? To clarify: I don't mind the prompt change in this PR, I'm just worried it hides a more serious problem

A-Mahla: Oh yeah, you’re probably right, and I’ve already fixed it (2 producers issues). I’ll retest without the prompt since I haven’t seen this bug in a while.

A-Mahla: I've tested without the prompt and it seems work well. One tool is call at a time :) So perfect don't need it

PR #250, opened 2026-03-12
249 cedar is the only available voice in the gui

RemiFabre:
## Fix: show all available voices in GUI instead of only Cedar

The voice dropdown was stuck on "Cedar" because of two issues:

1. **Gradio mode**: the dropdown was initialized with `choices=["cedar"]` and relied on an async callback to populate the rest. If the handler wasn't ready, it stayed stuck.
2. **Headless mode + frontend**: every error/fallback path in the `/voices` endpoint and JS `getVoices()` returned only `["cedar"]` instead of the full curated list.

### Changes
- `gradio_personality.py`: initialize voice dropdown with all 7 voices
- `headless_personality_ui.py`: return full voice fallback list in all error paths
- `static/main.js`: same for the frontend fallback

### DIFF
--- a/src/reachy_mini_conversation_app/gradio_personality.py
+++ b/src/reachy_mini_conversation_app/gradio_personality.py
@@ -10,7 +10,7 @@ from pathlib import Path
 
 import gradio as gr
 
-from .config import AVAILABLE_VOICES, LOCKED_PROFILE, config
+from .config import LOCKED_PROFILE, AVAILABLE_VOICES, config
 
 
 class PersonalityUI:
--- a/src/reachy_mini_conversation_app/headless_personality_ui.py
+++ b/src/reachy_mini_conversation_app/headless_personality_ui.py
@@ -13,7 +13,7 @@ from typing import Any, Callable, Optional
 
 from fastapi import FastAPI
 
-from .config import AVAILABLE_VOICES, LOCKED_PROFILE, config
+from .config import LOCKED_PROFILE, AVAILABLE_VOICES, config
 from .openai_realtime import OpenaiRealtimeHandler
 from .headless_personality import (
     DEFAULT_OPTION,


src/reachy_mini_conversation_app/static/main.js
```diff
@@ -163,6 +163,8 @@ async function applyPersonality(name, { persist = false } = {}) {
   return await resp.json();
 }
 
+const VOICE_FALLBACK = ["cedar", "alloy", "aria", "ballad", "verse", "sage", "coral"];
```
alozowski: the full list should be:
[
  "alloy",
  "ash",
  "ballad",
  "coral",
  "echo",
  "sage",
  "shimmer",
  "verse",
  "marin",
  "cedar"
]

`marin` and `cedar` are recommended for `gpt-realtime`, we can note it somewhere on this page

alozowski: the source: https://developers.openai.com/api/docs/guides/text-to-speech/#voice-options

src/reachy_mini_conversation_app/gradio_personality.py
```diff
@@ -108,7 +108,8 @@ def create_components(self) -> None:
         self.person_name_tb = gr.Textbox(label="Personality name", interactive=not is_locked)
         self.person_instr_ta = gr.TextArea(label="Personality instructions", lines=10, interactive=not is_locked)
         self.tools_txt_ta = gr.TextArea(label="tools.txt", lines=10, interactive=not is_locked)
-        self.voice_dropdown = gr.Dropdown(label="Voice", choices=["cedar"], value="cedar", interactive=not is_locked)
+        _initial_voices = ["cedar", "alloy", "aria", "ballad", "verse", "sage", "coral"]
```
alozowski: we also need to fix this list of available voices according to my previous comment

src/reachy_mini_conversation_app/headless_personality_ui.py
```diff
@@ -268,20 +268,31 @@ async def _do_apply() -> str:
         except Exception as e:
             return JSONResponse({"ok": False, "error": str(e)}, status_code=500)  # type: ignore
 
+    # Curated fallback matching OpenaiRealtimeHandler.get_available_voices()
+    _VOICE_FALLBACK: list[str] = [
```
alozowski: at this point, can we set up the list of voices only in one place for `.py` scripts and import it all the time we need it?

Issue #228, opened 2026-03-03
Update deprecated/non-existent OpenAI Realtime events


A-Mahla:
## Problem

`src/reachy_mini_conversation_app/openai_realtime.py` currently handles several OpenAI Realtime events that are deprecated or no longer valid.  
The project currently resolves `openai` to `2.21.0` in `uv.lock`, so event compatibility should be aligned to that target (or updated) and protected from future dependency drift.

Affected code area: [`src/reachy_mini_conversation_app/openai_realtime.py`](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/src/reachy_mini_conversation_app/openai_realtime.py#L309C1-L505C26).

### Deprecated / inexistent from `2.21.0` and latest `2.24.0`

- `conversation.item.input_audio_transcription.partial`
- `response.audio.completed`

### Not from Realtime protocol type (needs decision)

- `response.completed`  
  Question: do we still need this event?

### Preferred migration direction according the Openai documentation

Preferred event:

- `conversation.item.input_audio_transcription.partial` to `conversation.item.input_audio_transcription.delta`
- `response.audio_transcript.done` to `response.output_audio_transcript.done`
- `response.audio.done` to `response.output_audio.done`
- `response.audio.delta` to `response.output_audio.delta`

## Dependency stability concern

In `pyproject.toml`, the dependency is currently:

- `openai>=2.1`

The version currently targeted by lock resolution is:

- `uv.lock`: `name = "openai"` / `version = "2.21.0"`

Even if `uv.lock` is currently pinning resolved versions, this open-ended constraint can still introduce breaking changes when lock files are refreshed.

## Proposed action

- Upgrade to `2.24.0`.
- Tighten and maintain the `openai` version in `pyproject.toml` (for example: fixed version or controlled compatible range).
- Keep `uv.lock` updated.
- Migrate/remove deprecated/inexistent events.
- Strong typing of event models.

Issue #229, opened 2026-03-04
Partial event transcript logic has no reset and no accumulation


A-Mahla:
**File:** `src/reachy_mini_conversation_app/openai_realtime.py`

## Context

The [`conversation.item.input_audio_transcription.partial`](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/src/reachy_mini_conversation_app/openai_realtime.py#L350C1-L368C22) event code handles chunked transcription by tracking a sequence counter, `partial_transcript_sequence`.

#### ⚠️ To resove this, the issue https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/228 must be resolve because `conversation.item.input_audio_transcription.partial` is no longer support and couldn't be tested. First migrate to `conversation.item.input_audio_transcription.delta`

---

## 1 (fix): `partial_transcript_sequence` is never reset (fix)

The counter is initialized once in [`__init__`](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/src/reachy_mini_conversation_app/openai_realtime.py#L91C9-L91C50)  and [incremented on every delta event](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/src/reachy_mini_conversation_app/openai_realtime.py#L354), but **never reset**:

- Not when [`transcription.completed`](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/src/reachy_mini_conversation_app/openai_realtime.py#L371C2-L383C1) fires, the natural boundary between utterances.
- Not on session restart via `_restart_session()`.
- Not on `shutdown()`.

The counter grows unboundedly across the entire session lifetime. This makes the sequence semantically meaningless across utterances and prevents any future logic from reasoning about per-utterance sequence numbers.

**Fix:** Reset `self.partial_transcript_sequence = 0` when `transcription.completed` fires.

---

## 2 (logic improvement): Only the latest delta fragment is emitted (no item_id tracking or transcript reconstruction)

[...]

PR #251, opened 2026-03-13
Update deprecated/non-existent OpenAI Realtime events

A-Mahla:
## Summary

Resolve issue https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/228 (and sub-issue https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/229):

- Upgrade `openai` package to 2.24.0. (`uv.lock` and `pyproject.py` updated).
- Migrate or remove deprecated/nonexistent events.
- Strong typing for event models.

## Category
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic

### Run modes
- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)

### Vision / motion
- [ ] Head wobble

### Dependencies & config

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "python-dotenv",
 
     #OpenAI
-    "openai==2.24.0",
+    "openai==2.28.0",
 
     #Reachy mini
     "reachy_mini_dances_library",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "python-dotenv",
 
     #OpenAI
-    "openai>=2.1",
+    "openai==2.24.0",
 
     #Reachy mini
     "reachy_mini_dances_library",
--- a/src/reachy_mini_conversation_app/openai_realtime.py
+++ b/src/reachy_mini_conversation_app/openai_realtime.py
@@ -16,16 +16,18 @@ from fastrtc import AdditionalOutputs, AsyncStreamHandler, wait_for_item, audio_
 from pydantic import Field, BaseModel
 from numpy.typing import NDArray
 from scipy.signal import resample
-from openai.types.realtime import RealtimeSessionCreateRequestParam
+from openai.types.realtime import (
+    ConversationItemParam,
+    AudioTranscriptionParam,
+    RealtimeAudioConfigParam,
+    RealtimeAudioConfigInputParam,
+    RealtimeAudioConfigOutputParam,
+    RealtimeResponseCreateParamsParam,
+    RealtimeSessionCreateRequestParam,
+)
 from websockets.exceptions import ConnectionClosedError
 from openai.resources.realtime.realtime import AsyncRealtimeConnection
-from openai.types.realtime.conversation_item_param import ConversationItemParam
-from openai.types.realtime.audio_transcription_param import AudioTranscriptionParam
-from openai.types.realtime.realtime_audio_config_param import RealtimeAudioConfigParam
 from openai.types.realtime.realtime_audio_formats_param import AudioPCM
-from openai.types.realtime.realtime_audio_config_input_param import RealtimeAudioConfigInputParam
-from openai.types.realtime.realtime_audio_config_output_param import RealtimeAudioConfigOutputParam
-from openai.types.realtime.realtime_response_create_params_param import RealtimeResponseCreateParamsParam
 from openai.types.realtime.realtime_audio_input_turn_detection_param import ServerVad
 
 from reachy_mini_conversation_app.config import AVAILABLE_VOICES, config
@@ -568,10 +570,6 @@ class OpenaiRealtimeHandler(AsyncStreamHandler):
                             input_transcript.deltas.append(delta)
 
                         sequence_counter = len(input_transcript.deltas) - 1
-
-                        logger.info(f"Input transcript: {input_transcript.deltas}")
-                        logger.info(f"Item ID: {item_id}")
-                        logger.info(f"Sequence counter: {sequence_counter}")
 
                         # Cancel previous debounce task if it exists
                         if self.partial_transcript_task and not self.partial_transcript_task.done():
--- a/src/reachy_mini_conversation_app/openai_realtime.py
+++ b/src/reachy_mini_conversation_app/openai_realtime.py
@@ -22,7 +22,6 @@ from openai.resources.realtime.realtime import AsyncRealtimeConnection
 from openai.types.realtime.conversation_item_param import ConversationItemParam
 from openai.types.realtime.audio_transcription_param import AudioTranscriptionParam
 from openai.types.realtime.realtime_audio_config_param import RealtimeAudioConfigParam
-from openai.types.realtime.realtime_tools_config_param import RealtimeToolsConfigParam
 from openai.types.realtime.realtime_audio_formats_param import AudioPCM
 from openai.types.realtime.realtime_audio_config_input_param import RealtimeAudioConfigInputParam
 from openai.types.realtime.realtime_audio_config_output_param import RealtimeAudioConfigOutputParam
@@ -481,7 +480,7 @@ class OpenaiRealtimeHandler(AsyncStreamHandler):
                             voice=get_session_voice(),
                         ),
                     ),
-                    tools=cast(RealtimeToolsConfigParam, get_tool_specs()),
+                    tools=get_tool_specs(), # type: ignore[typeddict-item]
                     tool_choice="auto",
                 )
                 await conn.session.update(session=session_config)
--- a/src/reachy_mini_conversation_app/openai_realtime.py
+++ b/src/reachy_mini_conversation_app/openai_realtime.py
@@ -3,20 +3,28 @@ import base64
 import random
 import asyncio
 import logging
-from typing import Any, Final, Tuple, Literal, Optional
+from typing import Any, Final, Tuple, Literal, Optional, cast
 from pathlib import Path
 from datetime import datetime
 
 import cv2
 import numpy as np
 import gradio as gr
-from openai import AsyncOpenAI
 from fastrtc import AdditionalOutputs, AsyncStreamHandler, wait_for_item, audio_to_int16
-from openai.resources.realtime.realtime import AsyncRealtimeConnection
 from numpy.typing import NDArray
 from scipy.signal import resample
 from websockets.exceptions import ConnectionClosedError
 
+from openai import AsyncOpenAI
+from openai.types.realtime import RealtimeSessionCreateRequestParam
+from openai.resources.realtime.realtime import AsyncRealtimeConnection
+from openai.types.realtime.audio_transcription_param import AudioTranscriptionParam
+from openai.types.realtime.realtime_audio_config_param import RealtimeAudioConfigParam
+from openai.types.realtime.realtime_tools_config_param import RealtimeToolsConfigParam
+from openai.types.realtime.realtime_audio_formats_param import AudioPCM
+from openai.types.realtime.realtime_audio_config_input_param import RealtimeAudioConfigInputParam
+from openai.types.realtime.realtime_audio_config_output_param import RealtimeAudioConfigOutputParam
+from openai.types.realtime.realtime_audio_input_turn_detection_param import ServerVad
 from reachy_mini_conversation_app.config import config
 from reachy_mini_conversation_app.prompts import get_session_voice, get_session_instructions
 from reachy_mini_conversation_app.tools.core_tools import (
@@ -133,11 +141,15 @@ class OpenaiRealtimeHandler(AsyncStreamHandler):
             if self.connection is not None:
                 try:
                     await self.connection.session.update(
-                        session={
-                            "type": "realtime",
-                            "instructions": instructions,
-                            "audio": {"output": {"voice": voice}},
-                        },
+                        session=RealtimeSessionCreateRequestParam(
+                            type="realtime",
+                            instructions=instructions,
+                            audio=RealtimeAudioConfigParam(
+                                output=RealtimeAudioConfigOutputParam(
+                                    voice=voice,
+                                ),
+                            ),
+                        ),
                     )
                     logger.info("Applied personality via live update: %s", profile or "built-in default")
                 except Exception as e:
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "python-dotenv",
 
     #OpenAI
-    "openai>=2.1",
+    "openai==2.24.0",
```
alozowski: I think we don't need to lock the version, we can keep `"openai>=2.28.0"` ([new version is available](https://pypi.org/project/openai/#history))

openai/types/realtime/conversation_item_input_audio_transcription_delta_event.py
alozowski, line 1: It's a completely empty file, and I suggest to remove `openai/` directory fully

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -4,19 +4,30 @@
...
+from openai.resources.realtime.realtime import AsyncRealtimeConnection
+from openai.types.realtime.conversation_item_param import ConversationItemParam
+from openai.types.realtime.audio_transcription_param import AudioTranscriptionParam
+from openai.types.realtime.realtime_audio_config_param import RealtimeAudioConfigParam
+from openai.types.realtime.realtime_tools_config_param import RealtimeToolsConfigParam
+from openai.types.realtime.realtime_audio_formats_param import AudioPCM
+from openai.types.realtime.realtime_audio_config_input_param import RealtimeAudioConfigInputParam
+from openai.types.realtime.realtime_audio_config_output_param import RealtimeAudioConfigOutputParam
+from openai.types.realtime.realtime_response_create_params_param import RealtimeResponseCreateParamsParam
+from openai.types.realtime.realtime_audio_input_turn_detection_param import ServerVad
```
alozowski: These 11 deeply-nested type imports and 4 `cast()` calls add no runtime benefit: the `openai` lib accepts plain dicts at all these call sites, which is exactly how the lib authors write it themselves in their [README](https://github.com/openai/openai-python/blob/main/README.md#realtime-api) and [`push_to_talk_app.py`](https://github.com/openai/openai-python/blob/main/examples/realtime/push_to_talk_app.py) example

Please, revert to plain dicts, matching openai lib conventions. Keep only the `AsyncRealtimeConnection` type annotation for `self.connection` – that's a genuinely useful improvement over `Any`

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -361,11 +377,11 @@ async def _handle_tool_result(self, bg_tool: ToolNotification) -> None:
             # Send the tool result back
             if isinstance(bg_tool.id, str):
                 await self.connection.conversation.item.create(
-                    item={
+                    item=cast(ConversationItemParam, {
```
alozowski: `cast(ConversationItemParam, {...})` does zero runtime validation and it's strictly worse than a `# type: ignore[arg-type]` because it silences the checker while pretending correctness, so please, remove all `cast()` calls and pass plain dicts as before (the openai lib accepts them)

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -517,9 +526,7 @@ async def _run_realtime_session(self) -> None:
                         logger.debug("User speech stopped - server will auto-commit with VAD")
 
                     if event.type in (
-                        "response.audio.done",  # GA
-                        "response.output_audio.done",  # GA alias
-                        "response.audio.completed",  # legacy (for safety)
+                        "response.output_audio.done",  # GA
                         "response.completed",  # text-only completion
```
alozowski: `response.done` already covers both audio and text-only completions so we can remove `"response.completed"` as well

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -558,15 +571,17 @@ async def _run_realtime_session(self) -> None:
                             except asyncio.CancelledError:
                                 pass
 
-                        # Start new debounce timer with sequence number
+                        # Start new debounce timer with the last delta
                         self.partial_transcript_task = asyncio.create_task(
-                            self._emit_debounced_partial(event.transcript, current_sequence)
+                            self._emit_debounced_partial(chunks[-1], item_id, sequence_counter)
```
alozowski: `chunks[-1]` emits only the last delta fragment instead of the accumulated transcript. With the old `.partial event`, event.transcript was already the full partial but `.delta` is incremental. This should be `"".join(chunks)`

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -575,15 +590,18 @@ async def _run_realtime_session(self) -> None:
                             except asyncio.CancelledError:
                                 pass
 
+                        # Clear the accumulator entry for this item
+                        self.input_transcript_acc.pop(item_id, None)
```
alozowski: If `completed` is never received for an item (connection drops mid-utterance, API bug), this entry leaks for the rest of the session. The session-level `self.input_transcript_acc.clear()` at startup mitigates across restarts, but within a long session stale entries can accumulate

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -442,34 +458,24 @@ async def _run_realtime_session(self) -> None:
...
-                                    "type": "audio/pcm",
-                                    "rate": self.output_sample_rate,
-                                },
-                                "voice": get_session_voice(),
-                            },
-                        },
-                        "tools": get_tool_specs(),  # type: ignore[typeddict-item]
-                        "tool_choice": "auto",
-                    },
+                session_config = RealtimeSessionCreateRequestParam(
```
alozowski, line 469: I agree that this is more readable than the old deeply-nested dict, but it comes at the cost of eight fragile deep-submodule imports that can move between openai lib versions. Plus, openai lib authors use plain dicts for exactly this call in their own examples (cited in the previous comment). Please, revert to a well-indented dict

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -361,11 +377,11 @@ async def _handle_tool_result(self, bg_tool: ToolNotification) -> None:
             # Send the tool result back
             if isinstance(bg_tool.id, str):
                 await self.connection.conversation.item.create(
```
alozowski, line 387: I noticed that openai payload construction is repeated in several places, `in apply_personality`, three times in `_handle_tool_result`, and in `send_idle_signal`. You can leave a TODO for a next PR to create a couple of small helper functions so the shape lives in one place

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -95,8 +106,8 @@ def __init__(self, deps: ToolDependencies, gradio_mode: bool = False, instance_p
 
         # Debouncing for partial transcripts
         self.partial_transcript_task: asyncio.Task[None] | None = None
-        self.partial_transcript_sequence: int = 0  # sequence counter to prevent stale emissions
         self.partial_debounce_delay = 0.5  # seconds
+        self.input_transcript_acc: dict[str, list[str]] = {} # list of delta transcripts by item_id
```
alozowski: A minor naming suggestion to rename `input_transcript_acc` into `input_transcript_chunks_by_item`

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -4,19 +4,30 @@
 import random
 import asyncio
 import logging
-from typing import Any, Final, Tuple, Literal, Optional
+from typing import Any, Final, Tuple, Literal, Optional, cast
```
alozowski: Left a comment about cast (below) above

alozowski (changes requested):
I left some comments that are worth addressing

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -81,7 +92,7 @@ def __init__(self, deps: ToolDependencies, gradio_mode: bool = False, instance_p
         self.output_sample_rate = OPEN_AI_OUTPUT_SAMPLE_RATE
         self.input_sample_rate = OPEN_AI_INPUT_SAMPLE_RATE
 
-        self.connection: Any = None
+        self.connection: AsyncRealtimeConnection | None = None
```
alozowski, line 104: this is nice

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: I see your point, but I disagree. This isn’t only about runtime safety, it’s about readability and maintainability.

Using plain dicts make it hard for contributors to know what shape is expected by the OpenAI API. With typed models, the intent is explicit, and if the OpenAI SDK changes between versions, we’ll catch mismatches early via type checking (mypy), instead of discovering them at runtime.

pyproject.toml
A-Mahla: Yes, let’s go with `2.28.0`, but I’d prefer keeping a strict version to prevent breaking changes from being introduced on the user side.

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: The `cast()` calls don't catch mismatches but they hide them, that's literally what `cast()` does: it tells mypy to "not check this", we already stuck on this in ~November. So if the openai changes the shape, we won't get a type error, we'll get a silent runtime bug and it's the opposite of what you want

Plus, from my experience, imagine a dev that opens the openai examples and want to write exactly how they show, it's easier

Anyway, if you can find how to add typing without relying on cast + simplifier this block, it's acceptable

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Related to my comment here https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/251#discussion_r2940249162. `cast()` is pretty standard to use with union types. I think it’s more readable and informative, since it makes it clear which kind of message can be sent. This is a union type: even if we add more models to the union, the union name or intent is very unlikely to change. see https://github.com/openai/openai-python/blob/fc7f598efa70c6bef5ad15d1a8c9dfcd0fbd33fa/src/openai/types/realtime/conversation_item_param.py#L20

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: removed

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Changed 👍

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Answered above https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/251#discussion_r2940249162 and for openai version drifting https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/251#discussion_r2940451578

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Answered above https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/251#discussion_r2940614683

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Changed 👍

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: I've changed the logic to handle one `item_id` at a time, which overwrites the previous accumulation list. So there are no more accumulation leaks.

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: Added 👍

src/reachy_mini_conversation_app/openai_realtime.py
A-Mahla: I understand what `cast()` implies, but I’m using it only for a union type, as explained here: https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/251#discussion_r2940614683. If, after reading that, you still think it’s a problem, I’m open to defining the exact type it's referring to within the union. 👍

A-Mahla: I've simply the import block  https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/8a1f432bb66f8bd73e2e15a68cf6c10866283dc6/src/reachy_mini_conversation_app/openai_realtime.py#L19-L27

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -95,8 +116,8 @@ def __init__(self, deps: ToolDependencies, gradio_mode: bool = False, instance_p
 
         # Debouncing for partial transcripts
         self.partial_transcript_task: asyncio.Task[None] | None = None
-        self.partial_transcript_sequence: int = 0  # sequence counter to prevent stale emissions
         self.partial_debounce_delay = 0.5  # seconds
+        self.input_transcript_chunks_by_item = InputTranscriptChunksByItem()
 
         # Internal lifecycle flags
         self._shutdown_requested: bool = False
```
alozowski: seems like we don't use `_shutdown_requested` at all

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -558,15 +579,17 @@ async def _run_realtime_session(self) -> None:
...
                         self.partial_transcript_task = asyncio.create_task(
-                            self._emit_debounced_partial(event.transcript, current_sequence)
+                            self._emit_debounced_partial("".join(input_transcript.deltas), item_id, sequence_counter)
                         )
 
                     # Handle completed transcription (user finished speaking)
                     if event.type == "conversation.item.input_audio_transcription.completed":
                         logger.debug(f"User transcript: {event.transcript}")
 
+                        item_id = event.item_id
```
alozowski: seems like it is assigned but never used

alozowski (approved):

Issue #258, opened 2026-03-17
Modernise the local vision pipeline and cut unnecessary memory use


alozowski:
The local vision path still uses older loading and preprocessing patterns that do extra work and waste memory. We should update it to the recent Transformers API and simplify how images and generated tokens are handled. This removes the local JPEG and base64 roundtrip, avoids repeated input preparation, and improves OOM recovery behavior. The goal is to keep the same feature while making inference lighter and more robust. It also leaves the vision code easier to maintain.

PR #259, opened 2026-03-18
Modernise local vision

alozowski:
## Summary
Solves https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/258. The local vision feature is lighter, simpler, and more robust. It keeps the same local vision feature, but removes older preprocessing and lifecycle patterns that were doing extra work and increasing memory pressure.

Changes sorted by their importance:
- updates the optional local vision dependency stack to `transformers==5.3.0` and adds `accelerate>=1.13.0`;
- updates local vision inference to the recent Transformers API, including direct PIL image inputs through `apply_chat_template`, device-aware tensor transfer, and decoding only newly generated tokens instead of post-processing the full generated prompt;
- replaces the old background `VisionManager` flow with an on-demand `VisionProcessor`, so local vision only runs when the camera tool is actually invoked instead of processing frames periodically in the background;
- removes the local jpeg and base64 roundtrip in the local vision path, which cuts unnecessary image copies and preprocessing overhead before inference;
- simplifies the camera tool wiring by passing a `vision_processor` directly through tool dependencies and using the local processor only when `--local-vision` is enabled;
- fixes the vision prompt;
- improves failure handling during local vision startup by failing fast when imports are missing or crash at import time;
- adds related updates in README and CLI wording
 
## Category
- [x] Fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [x] Headless mode (default)
- [x] Gradio UI (`--gradio`)
- [x] Simulation (`--gradio` required)
</details>

<details>

- [x] Local vision (`--local-vision`)
- [x] Head wobble
</details>

## Notes
I noticed that movement stops during ongoing speech after tool calls in realtime conversation loop – can be fixed in a separate PR

### DIFF
--- a/src/reachy_mini_conversation_app/utils.py
+++ b/src/reachy_mini_conversation_app/utils.py
@@ -25,7 +25,7 @@ def parse_args() -> tuple[argparse.Namespace, list]:  # type: ignore
         "--head-tracker",
         choices=["yolo", "mediapipe"],
         default=None,
-        help="Choose the head-tracking backend: yolo uses a local face detector, mediapipe uses reachy_mini_toolbox",
+        help="Head-tracking backend: yolo uses a local face detector, mediapipe uses reachy_mini_toolbox. Disabled by default.",
     )
     parser.add_argument("--no-camera", default=False, action="store_true", help="Disable camera usage")
     parser.add_argument(


src/reachy_mini_conversation_app/vision/processors.py
```diff
@@ -60,272 +62,140 @@ def _determine_device(self) -> str:
...
     ) -> str:
-        """Process CV2 image and return description with retry logic."""
+        """Process a BGR camera frame and return a text description."""
         if not self._initialized or self.processor is None or self.model is None:
             return "Vision model not initialized"
 
+        processor = self.processor
+        model = self.model
+        rgb_image = Image.fromarray(np.ascontiguousarray(frame[..., ::-1]))
+        request = prompt.strip() or DEFAULT_LOCAL_VISION_PROMPT
```
andimarafioti: This or doesn't do anything. If there was no prompt, the default for the method is already DEFAULT_LOCAL_VISION_PROMPT

src/reachy_mini_conversation_app/vision/processors.py
```diff
@@ -1,35 +1,37 @@
...
 
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_LOCAL_VISION_PROMPT = (
+    "Respond to the request using only details that are clearly visible in the image. "
+    "Do not guess, infer hidden details, or invent missing information. "
+    "If the answer is not clearly visible, say exactly: I can't tell from this image. "
+    "Keep the answer short and factual."
+)
```
andimarafioti: Is this ever used? It seems to me like this is some steering you want to do, and it should be added to the image prompt, not put instead of the prompt. Further, if there is no question, then this default prompt doesn't help at all.

src/reachy_mini_conversation_app/main.py
```diff
@@ -105,7 +105,11 @@ def run(
         logger.info("Simulation mode detected. Automatically enabling gradio flag.")
         args.gradio = True
 
-    camera_worker, _, vision_manager = handle_vision_stuff(args, robot)
+    try:
+        camera_worker, vision_processor = initialize_camera_and_vision(args, robot)
+    except (ImportError, RuntimeError) as e:
+        logger.error("Failed to initialize camera/vision: %s", e)
+        sys.exit(1)
```
andimarafioti, line 113: What are you silently letting pass here? If the initialize fails with a different error, this will silently catch it. Do we want that?

src/reachy_mini_conversation_app/utils.py
```diff
@@ -1,20 +1,27 @@
...
+import subprocess
+from typing import TYPE_CHECKING, Optional
 
 from reachy_mini import ReachyMini
 from reachy_mini_conversation_app.camera_worker import CameraWorker
 
 
-def parse_args() -> Tuple[argparse.Namespace, list]:  # type: ignore
+if TYPE_CHECKING:
+    from reachy_mini_conversation_app.vision.processors import VisionProcessor
```
andimarafioti, line 14: Is this a ruff thing?

src/reachy_mini_conversation_app/utils.py
```diff
@@ -1,20 +1,27 @@
...
+def parse_args() -> tuple[argparse.Namespace, list]:  # type: ignore
     """Parse command line arguments."""
     parser = argparse.ArgumentParser("Reachy Mini Conversation App")
     parser.add_argument(
         "--head-tracker",
-        choices=["yolo", "mediapipe", None],
+        choices=["yolo", "mediapipe"],
         default=None,
-        help="Choose head tracker (default: None)",
+        help="Choose the head-tracking backend: yolo uses a local face detector, mediapipe uses reachy_mini_toolbox",
```
andimarafioti: You could add "Head tracking is disabled by default"

src/reachy_mini_conversation_app/utils.py
```diff
@@ -34,18 +41,16 @@ def parse_args() -> Tuple[argparse.Namespace, list]:  # type: ignore
...
-    camera_worker = None
+def initialize_camera_and_vision(
+    args: argparse.Namespace,
+    current_robot: ReachyMini,
+) -> tuple[CameraWorker | None, VisionProcessor | None]:
+    """Initialize camera capture, optional head tracking, and optional local vision."""
+    camera_worker: CameraWorker | None = None
     head_tracker = None
-    vision_manager = None
+    vision_processor: VisionProcessor | None = None
```
andimarafioti:
```suggestion
    vision_processor: Optional[VisionProcessor] = None
```

src/reachy_mini_conversation_app/utils.py
```diff
@@ -34,18 +41,16 @@ def parse_args() -> Tuple[argparse.Namespace, list]:  # type: ignore
...
-    By default, vision is handled by gpt-realtime model when camera tool is used.
-    If --local-vision flag is used, a local vision model will process images periodically.
-    """
-    camera_worker = None
+def initialize_camera_and_vision(
+    args: argparse.Namespace,
+    current_robot: ReachyMini,
+) -> tuple[CameraWorker | None, VisionProcessor | None]:
+    """Initialize camera capture, optional head tracking, and optional local vision."""
+    camera_worker: CameraWorker | None = None
```
andimarafioti:
```suggestion
    camera_worker: Optional[CameraWorker] = None
```

src/reachy_mini_conversation_app/vision/processors.py
alozowski: I see why I got lost in the prompting, I tried to remove a basic one we had before – will fix

src/reachy_mini_conversation_app/main.py
alozowski: You're right!

src/reachy_mini_conversation_app/utils.py
alozowski: Yes, ruff at its finest

PR #264, opened 2026-03-20
Shorten built-in profile paths for Windows

andimarafioti:
## Summary
- move the built-in profiles out of `src/reachy_mini_conversation_app/` into a top-level `profiles/` directory to reduce Windows path length
- keep the built-in profile layout simple by using direct profile names with `instructions.txt` and `tools.txt`
- add regression coverage for profile loading, collision detection, and a project-wide Windows path budget of `104` characters

## Testing
- `uv run ruff check src/reachy_mini_conversation_app/config.py src/reachy_mini_conversation_app/headless_personality.py src/reachy_mini_conversation_app/gradio_personality.py src/reachy_mini_conversation_app/tools/core_tools.py tests/test_profile_paths.py tests/test_config_name_collisions.py tests/test_external_loading.py tests/test_openai_realtime.py`
- `uv run pytest tests/test_profile_paths.py tests/test_config_name_collisions.py tests/test_external_loading.py`
- `uv run --extra local_vision pytest`

## Notes
- the longest checked-in project path is now `70` characters relative to the repo root, which is comfortably below the `104` character budget from review
- `uv run pytest` in the base environment still is not the right full-suite command because `tests/vision/test_processors.py` imports `torch`; the full suite passes with the `local_vision` extra enabled

### DIFF
--- a/tests/test_profile_paths.py
+++ b/tests/test_profile_paths.py
@@ -14,9 +14,31 @@ from reachy_mini_conversation_app.headless_personality import (
 )
 
 
+# Path characters budget computation
+# ─────────────────
+# Windows MAX_PATH limit: 259 usable characters (failures start at 260)
+#
+# Project files (WINDOWS_PATH_BUDGET = 130):
+#   C:\Users\<username(20)>
+#     \.cache\huggingface\hub
+#     \spaces--pollen-robotics--reachy_mini_conversation_app
+#     \snapshots\<commit_hash(40)>\
+#   = 158 characters  =>  101 remaining to 259.
+#   The project root folder is not cloned in the snapshot, so we add it
+#   back to the budget: 101 + len("reachy_mini_conversation_app\") (29) = 130.
+#
+# Wheel files (WINDOWS_WHEEL_PATH_BUDGET = 71):
+#   C:\Users\<username(20)>
+#     \.cache\huggingface\hub
+#     \spaces--pollen-robotics--reachy_mini_conversation_app
+#     \snapshots\<commit_hash(40)>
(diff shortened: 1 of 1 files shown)


tests/test_openai_realtime.py
```diff
@@ -590,7 +590,9 @@ async def _fake_dispatch(
         )
 
     # Yield so spawned tool tasks, the listener, and the sender can drain.
-    await asyncio.sleep(5)
+    # This stress test queues hundreds of serialized response.create calls, so
+    # slower CI runners need a wider drain window before teardown.
```
andimarafioti, line 594: who are you calling a slower ci runner!

tests/test_profile_paths.py
```diff
@@ -0,0 +1,51 @@
...
+    project_files = [
+        path
+        for path in project_root.rglob("*")
+        if path.is_file() and not any(part in ignored_parts for part in path.relative_to(project_root).parts)
+    ]
+
+    longest_path = max(project_files, key=lambda path: len(str(path.relative_to(project_root))))
+    longest_length = len(str(longest_path.relative_to(project_root)))
+
+    assert longest_length <= 140, (
```
CarolinePascal: Looks like Claude took the absolute path to compute the 140 character value, I'll compute a better estimate asap !

CarolinePascal: At the moment, the longest path is 118 characters, which will lead to 273 characters on Windows with a username of 20 characters (maximum possible value).

We need to remove 14 characters to stay under the 260 characters limit, so the maximum size should be 104 characters 😬

PR #271, opened 2026-03-25
fix(resolve): inverting parents/resolve for all paths in the codebase

CarolinePascal:
Follow up of #270


alozowski (approved):
These changes indeed make sense in symlinked/junction-based setups

PR #307, opened 2026-04-22
Preserve selected voice when changing profiles

### LINKED ISSUE #304: Changing profiles resets voice
When I select a voice in the UI and apply it, it changes the voice to the one I selected. If later I change the profile, the UI still shows the voice I had selected, but the robot speaks in the default voice for the backend selected.

andimarafioti:
## Summary
Keep the manually selected voice active when the user changes profiles from the settings UI.

## Root Cause
Applying a profile called `apply_personality()` in both the OpenAI Realtime and Gemini Live handlers. That path was resetting the handler's in-memory voice override, so the runtime session restarted with the profile/default backend voice even though the UI still displayed the previously selected voice.

## What Changed
- Stopped clearing the handler voice override when applying a personality.
- Made the OpenAI personality-apply path prefer the existing manual voice override when sending the live session update.
- Added backend-level regressions for OpenAI and Gemini covering: apply voice, then change profile, and confirm the selected voice is preserved.

## Impact
Changing profiles no longer silently resets the robot speech voice back to the backend default while the UI still shows the user-selected voice.

## Validation
- `ruff check src/reachy_mini_conversation_app/openai_realtime.py src/reachy_mini_conversation_app/gemini_live.py tests/test_openai_realtime.py tests/test_gemini_live.py`
- `uv run pytest -q tests/test_openai_realtime.py::test_apply_personality_preserves_manual_voice_override tests/test_gemini_live.py::test_apply_personality_preserves_manual_voice_override`

Closes #304

### DIFF
--- a/src/reachy_mini_conversation_app/gemini_live.py
+++ b/src/reachy_mini_conversation_app/gemini_live.py
@@ -184,7 +184,12 @@ class GeminiLiveHandler(AsyncStreamHandler):
 
     def copy(self) -> "GeminiLiveHandler":
         """Create a copy of the handler."""
-        return GeminiLiveHandler(self.deps, self.gradio_mode, self.instance_path)
+        return GeminiLiveHandler(
+            self.deps,
+            self.gradio_mode,
+            self.instance_path,
+            startup_voice=self._voice_override,
+        )
 
     def _set_listening_state(self, listening: bool) -> None:
         """Avoid queueing redundant listening-state updates."""
--- a/src/reachy_mini_conversation_app/gemini_live.py
+++ b/src/reachy_mini_conversation_app/gemini_live.py
@@ -30,7 +30,6 @@ from reachy_mini_conversation_app.config import (
     GEMINI_AVAILABLE_VOICES,
     DEFAULT_VOICE_BY_BACKEND,
     config,
-    get_persisted_voice_override,
 )
 from reachy_mini_conversation_app.prompts import get_session_voice, get_session_instructions
 from reachy_mini_conversation_app.tools.core_tools import (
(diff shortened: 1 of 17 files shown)


alozowski:
I agree with the live profile-change fix: applying a profile should keep the current in-memory voice override, but I have a concern with the startup persistence part: I'd avoid adding a new `REACHY_MINI_VOICE_OVERRIDE` env var because the selected voice is UI startup state, not an environment configuration. I think a cleaner option could be adding a small `startup_settings.py` module that reads/writes an instance-local `startup_settings.json` with the saved startup profile and voice. This keeps `.env` focused on operational config like API keys, backend/model, and external paths, so we don't need to expose or document a new voice env var. It also avoids exposing handler internals via `get_voice_override()`, since the settings route can persist the actual behavior via `get_current_voice()`

andimarafioti:
@alozowski addressed in `5029bd2`.

I reworked the startup persistence path to match your suggestion:
- removed the `REACHY_MINI_VOICE_OVERRIDE` env var approach
- added instance-local startup settings storage in `startup_settings.json`
- kept `.env` for operational config only
- removed `get_voice_override()` and persist the current behavior via `get_current_voice()` in the settings route

The live profile-change fix is unchanged; this only replaces how the saved startup profile/voice state is persisted and restored.

src/reachy_mini_conversation_app/startup_settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def load_startup_settings_into_runtime(instance_path: str | Path | None) -> StartupSettings:
+    """Load instance-local startup settings when no explicit profile override is set."""
+    from reachy_mini_conversation_app.config import LOCKED_PROFILE, set_custom_profile
+
+    if LOCKED_PROFILE is not None:
+        return StartupSettings()
+    if os.getenv("REACHY_MINI_CUSTOM_PROFILE"):
+        return StartupSettings()
```
alozowski: If the instance `.env` already contains `REACHY_MINI_CUSTOM_PROFILE`, we return empty startup settings and never restore the saved voice from `startup_settings.json`. I think env should only win for the profile value, not prevent loading the persisted startup voice.

src/reachy_mini_conversation_app/openai_realtime.py
alozowski: This copy doesn't carry the restored `_voice_override`, but `fastrtc` creates per-session handlers via `copy()`. That means startup voice restoration can work on the original handler but still be lost in the actual streaming session

src/reachy_mini_conversation_app/gemini_live.py
alozowski: Same as for OpenaiRealtimeHandler

alozowski (changes requested):
The updated scope looks much better now, thank you! I'd only fix the `.env` fallback interaction and make sure copied stream handlers preserve the restored startup voice

src/reachy_mini_conversation_app/startup_settings.py
andimarafioti: Addressed in `43c0cc0`.

`load_startup_settings_into_runtime()` now still reads `startup_settings.json` when `.env` provides `REACHY_MINI_CUSTOM_PROFILE`, but only lets the env value win for profile selection. The saved startup voice is still returned and passed into the handler on boot. I also updated the startup-settings regression to cover that case.

src/reachy_mini_conversation_app/openai_realtime.py
andimarafioti: Addressed in `43c0cc0`.

`OpenaiRealtimeHandler.copy()` now forwards the current `_voice_override` into the copied handler, so the per-session `fastrtc` handler keeps the restored startup voice. I also added a regression test that exercises the copy path explicitly.

src/reachy_mini_conversation_app/gemini_live.py
andimarafioti: Addressed in `43c0cc0`.

Same fix here: `GeminiLiveHandler.copy()` now carries the current `_voice_override` into the copied handler, and there is a matching regression test for the Gemini copy path as well.

alozowski (approved):
let's go!

Issue #301, opened 2026-04-21
Do not expose head_tracking tool unless head tracking is enabled at runtime


alozowski:
The `head_tracking` tool is currently loaded from profile `tools.txt` files and exposed to the assistant even when the app is started without `--head-tracker`. This means a normal run like `reachy-mini-conversation-app` can still let the model call `head_tracking`, even though no head-tracking backend was initialised.

This creates a misleading runtime state: the tool may return "head tracking started" while no actual tracking can happen, or while camera/head-tracker dependencies are unavailable.

Issue #306, opened 2026-04-22
Investigate flaky yolo_process timeout-drain test on macOS and Windows


andimarafioti:
## Summary
`pytest` is failing on macOS and Windows for a platform-sensitive `yolo_process` test that appears unrelated to PR #305.

Failing test:
- `tests/vision/test_yolo_process.py::test_head_tracker_skips_new_frame_until_timed_out_reply_is_drained`

## Current CI context
On PR #305 (`ea4ef83b7d90abe958fc342fb705cb6538a69113`):
- Ubuntu pytest passed
- macOS pytest failed
- Windows pytest failed
- `ruff` and `mypy` passed

Workflow run:
- https://github.com/pollen-robotics/reachy_mini_conversation_app/actions/runs/24766840161

Example failing jobs:
- macOS: https://github.com/pollen-robotics/reachy_mini_conversation_app/actions/runs/24766840161/job/72463053310
- Windows: https://github.com/pollen-robotics/reachy_mini_conversation_app/actions/runs/24766840161/job/72463053305

## Failure details
The test expects a delayed reply to be drained and the next request to succeed after a short sleep:

```py
tracker = YoloHeadTrackerProcess(request_timeout=0.01)
...
time.sleep(0.15)
eye_center, roll = tracker.get_head_position(frame)
assert eye_center is not None
```

Observed failures on both macOS and Windows:
- `assert None is not None`
- warning log: `Head tracker yolo communication timed out: yolo head tracker timed out after 0.01s`

## Why this looks separate from #305
The PR only changes profile UI loading and tests around default profile tools. Nothing in that PR touches:
- `tests/vision/test_yolo_process.py`
- `vision/head_tracking/yolo_process.py`
- subprocess/timing behavior

The latest `main` Pytest workflow is green, so this currently looks like a flaky or platform-sensitive test rather than a deterministic regression from the profile UI work.

## Suspected area
Either:
- the test is too timing-sensitive for macOS/Windows scheduler/process timing, or
- `YoloHeadTrackerProcess` has a real cross-platform timing edge case in delayed reply draining.

## Suggested next step
Reproduce by repeatedly running:

```bash
pytest -q tests/vision/test_yolo_process.py::test_head_tracker_skips_new_frame_until_timed_out_reply_is_drained
```

Preferably on macOS and Windows, then decide whether to harden the test or adjust the timeout-drain logic in `vision/head_tracking/yolo_process.py`.

PR #320, opened 2026-04-22
Fix: hide head tracking when disabled

yaseminozkut:
## Summary

- Fixes issue #301 
- When `--head-tracker` is not set (or `--no-camera` is used), the head_tracking tool is excluded from the tool list sent to both OpenAI Realtime and Gemini Live
- Added startup logging for `--head-tracker` enabled/disabled and active tool set logging when user starts the conversation session 

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [x] Head tracker (`--head-tracker {yolo,mediapipe}`)
- [x] Camera pipeline (with/without `--no-camera`)
- [ ] Head wobble
</details>

### DIFF
--- a/src/reachy_mini_conversation_app/config.py
+++ b/src/reachy_mini_conversation_app/config.py
@@ -237,7 +237,6 @@ class Config:
     TOOLS_DIRECTORY = Path(_tools_directory_env) if _tools_directory_env else None
     AUTOLOAD_EXTERNAL_TOOLS = _env_flag("AUTOLOAD_EXTERNAL_TOOLS", default=False)
     REACHY_MINI_CUSTOM_PROFILE = LOCKED_PROFILE or os.getenv("REACHY_MINI_CUSTOM_PROFILE")
-    HEAD_TRACKER_ENABLED: bool = False
 
     logger.debug(f"Custom Profile: {REACHY_MINI_CUSTOM_PROFILE}")
 
@@ -345,11 +344,6 @@ def get_default_voice_for_backend(backend: str | None = None) -> str:
 def is_gemini_model() -> bool:
     """Return True if the configured MODEL_NAME is a Gemini Live model."""
     return get_backend_choice() == GEMINI_BACKEND
-
-
-def set_head_tracker_enabled(enabled: bool) -> None:
-    """Set whether a head-tracker backend was initialised at startup."""
-    config.HEAD_TRACKER_ENABLED = enabled
 
 
 def set_custom_profile(profile: str | None) -> None:
--- a/src/reachy_mini_conversation_app/config.py
+++ b/src/reachy_mini_conversation_app/config.py
@@ -237,6 +237,7 @@ class Config:
     TOOLS_DIRECTORY = Path(_tools_directory_env) if _tools_directory_env else None
     AUTOLOAD_EXTERNAL_TOOLS = _env_flag("AUTOLOAD_EXTERNAL_TOOLS", default=False)
     REACHY_MINI_CUSTOM_PROFILE = LOCKED_PROFILE or os.getenv("REACHY_MINI_CUSTOM_PROFILE")
+    HEAD_TRACKER_ENABLED: bool = False
 
     logger.debug(f"Custom Profile: {REACHY_MINI_CUSTOM_PROFILE}")
 
@@ -344,6 +345,11 @@ def get_default_voice_for_backend(backend: str | None = None) -> str:
 def is_gemini_model() -> bool:
     """Return True if the configured MODEL_NAME is a Gemini Live model."""
     return get_backend_choice() == GEMINI_BACKEND
+
+
+def set_head_tracker_enabled(enabled: bool) -> None:
+    """Set whether a head-tracker backend was initialised at startup."""
+    config.HEAD_TRACKER_ENABLED = enabled
 
 
 def set_custom_profile(profile: str | None) -> None:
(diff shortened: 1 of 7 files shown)


alozowski:
Hello @yaseminozkut!

Thank you for this PR – testing 🚀

src/reachy_mini_conversation_app/config.py
```diff
@@ -237,6 +237,7 @@ class Config:
     TOOLS_DIRECTORY = Path(_tools_directory_env) if _tools_directory_env else None
     AUTOLOAD_EXTERNAL_TOOLS = _env_flag("AUTOLOAD_EXTERNAL_TOOLS", default=False)
     REACHY_MINI_CUSTOM_PROFILE = LOCKED_PROFILE or os.getenv("REACHY_MINI_CUSTOM_PROFILE")
+    HEAD_TRACKER_ENABLED: bool = False
```
alozowski: I think this is the part I’d avoid. `head_tracking` availability is runtime capability (not config), and storing it as a process-global flag makes it easy to drift out of sync with the backend that was actually initialised. Could we derive this from the current session deps instead (`camera_worker` / `head_tracker`)?

src/reachy_mini_conversation_app/gemini_live.py
```diff
@@ -366,8 +366,12 @@ def _build_live_config(self) -> types.LiveConnectConfig:
         instructions = get_session_instructions()
         voice = _resolve_gemini_voice(self._voice_override or get_session_voice())
 
+        exclusion_list: list[str] = []
+        if not config.HEAD_TRACKER_ENABLED:
```
alozowski: This check is now duplicated across the realtime handlers and depends on global runtime state. I think the cleaner direction would be to have one shared path compute the available tool specs from the current deps, then let both handlers consume that same filtered tool list

src/reachy_mini_conversation_app/openai_realtime.py
```diff
@@ -499,6 +499,13 @@ async def _handle_tool_result(self, bg_tool: ToolNotification) -> None:
 
     async def _run_realtime_session(self) -> None:
         """Establish and manage a single realtime session."""
+        exclusion_list: list[str] = []
+        if not config.HEAD_TRACKER_ENABLED:
```
alozowski: Same concern as in the Gemini handler: this would be simpler and safer if tool availability came from one shared runtime-deps path instead of a duplicated global-config check here.

alozowski (changes requested):
It's a nice direction! I do think the implementation could stay a bit closer to the runtime state we already have, though. Right now this adds a process-global flag in `config` and repeats the same exclusion logic in both realtime handlers. That works, but it feels a little heavier than necessary and could drift from the backend that was actually initialised.

I think the cleaner path would be to derive `head_tracking` availability from the current session deps (`deps.camera_worker` / `deps.camera_worker.head_tracker`) and have both handlers consume the same filtered tool list from one shared place. That would keep `config` focused on static/env config, give us a single source of truth for tool availability, and remove the duplicated logic here.

I'd also love to see the regression pinned with a behavior-level test per backend, just to make sure we don't reintroduce this later

yaseminozkut:
Hi @alozowski !
Thank you for the detailed feedback! I'll take a look and implement what you suggested for removing the repetitive logic and global flag and also adding the tests for backend.

alozowski:
@yaseminozkut awesome! Feel free to ping me when you're ready for a re-review 👍🏻

yaseminozkut:
@alozowski thanks!!

yaseminozkut:
Hi @alozowski! I did the changes and added the tests for both backends as well. Is this the kind of implementation and tests you were thinking of? 

Also the flaky yolo test is I believe still flaky, it failed in CI. I believe it is the same issue from recent #306 (which is closed now) and unrelated to this PR.

alozowski (approved):
Thank you for your changes, looks good!

alozowski:
One small unrelated thing I noticed while re-running the tests: some of the sync Gemini handler tests are order-dependent because `GeminiLiveHandler` expects a current event loop at construction time, and `test_openai_realtime.py` clears the main-thread loop in one code path. I don't think that’s introduced by this PR, so I wouldn't block on it here, but it would be worth cleaning up separately to make those tests pass reliably in isolation too.

alozowski:
As for the YOLO test, could you please try merging the current main branch into your branch and check whether it still throws an error? I tried to resolve this issue one last time recently

yaseminozkut:
Hi @alozowski! I merged the main branch to my branch, seems like it solved, no issues with YOLO. Also for the Gemini tests, I can make `test_gemini_excludes_head_tracking_when_no_head_tracker` function introduced in this PR async. But the rest not related with this PR still remains to be fixed as well. Do you prefer me to make just make head tracking test function async as a partial fix for now? Or are you thinking of another general fix as another PR for all?

alozowski:
@yaseminozkut I think we're good to merge this as is now, so I'm merging! 

The implementation and tests here match what I had in mind, and I’d rather keep this PR focused on the head-tracking fix. If we still want to clean up the broader Gemini/event-loop test situation, I'd do that as a separate follow-up PR instead of mixing a partial fix into this one

Thank you for your work! 🫶

yaseminozkut:
Thank you as well @alozowski ! Happy to contribute 🫶🏼

PR #206, opened 2026-02-12
Feat: Adding support for MCPs in the conversion app

joakimeriksson:
## Summary
This PR adds the ability to connect MCP Servers to the Reachy-Mini conversation app. That way, it can support generic tools without explicitly adding new tool code in the code base. I have tested the control of my lamps and the readout of data from IoT devices in the Dirigera ecosystem (IKEA IoT). 

So far, mainly been tested in the simulator but connected to a real MCP server.

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

### Basic
- [x] CI green (Ruff, Tests, Mypy)
- [ ] Code update is clear (types, docs, comments)

### Run modes
- [x] Headless mode (default)
- [x] Gradio UI (`--gradio`)
- [x] Everything is tested in simulation as well (`--gradio` required)

### Vision / motion
- [x] Profiles or custom tools

### Dependencies & config
- [x] Updated `.env.example` if new config vars added

PR #322, opened 2026-04-23
Introduce MCP-based remote tool integration

### LINKED ISSUE #214: Integrate tool store

alozowski:
## Summary
This PR introduces MCP-based remote tool integration and de-scopes the previous [tool-store sync approach](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/222). Related to https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/214
It adds an MCP client for remote tool discovery and invocation, and a `tool-spaces add/list/remove` CLI to install public Hugging Face Spaces as remote tool sources. `add` installs and enables in the active profile by default, `--install-only` and `--profile <NAME>` override that

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [ ] Head wobble
</details>


alozowski:
🐥 Two canary spaces: 
- https://huggingface.co/spaces/alozowski/reachy-mini-search-tool
- https://huggingface.co/spaces/alozowski/reachy-mini-weather-tool

joakimeriksson:
How it this compared to my PR? Can I do the same things with this? It is a well-needed feature! So I am positive. Can I use secured MCPs with API tokens?

joakimeriksson:
Do you want me to test it on my Dirigera IoT control MCP server? (the one I tested my MCP client with).

alozowski:
Hello @joakimeriksson,

The key difference is that this PR scopes things to public Hugging Face Spaces only: install by slug, tools get properly namespaced, and they're still gated by each profile's `tools.txt`. [Your PR](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/206) is more flexible with arbitrary MCP URLs and inline tokens, but I wanted to start narrow and extend from there. Secured MCPs with API tokens aren't in this v1 yet, but that's definitely on the roadmap as a next step (=next PR)!

andimarafioti:
Hi @alozowski , how can we test this?

andimarafioti:
And codex says:
`  - src/reachy_mini_conversation_app/tool_spaces.py:103 and src/reachy_mini_conversation_app/tools/core_tools.py:311: installed Spaces are de-duped only by slug, but runtime matching uses the normalized alias prefix. Two valid slugs like
    owner/my-tool and owner/my_tool normalize to the same alias, so enabling a tool from one Space makes _resolve_remote_tools() try to satisfy it from both Spaces. That can fail startup with “tool unavailable” or produce duplicate tool
    names. The manifest read/install path should reject duplicate aliases or persist a disambiguated alias and honor it on read.

`

alozowski:
@andimarafioti this codex comment makes sense – I addressed it 
regarding the testing, I answered internally

src/reachy_mini_conversation_app/mcp_client.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def __init__(self, server: RemoteMcpServerConfig) -> None:
+        """Store one allowlisted server configuration and an in-memory tool cache."""
+        self.server = server
+        self._tool_index: dict[str, RemoteToolSpec] = {}
+
+    async def list_tool_specs(self) -> list[RemoteToolSpec]:
+        """Discover remote tools and translate them into app-facing specs."""
+        try:
+            async with self._session() as session:
```
FabienDanieau, line 284: self._session is not in init. Is it possible to open it there so we share a session for call_tools as well?

src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -131,8 +131,8 @@ def __init__(
         self.connection: AsyncRealtimeConnection | None = None
         self.output_queue: "asyncio.Queue[Tuple[int, NDArray[np.int16]] | AdditionalOutputs]" = asyncio.Queue()
 
-        self.last_activity_time = asyncio.get_event_loop().time()
-        self.start_time = asyncio.get_event_loop().time()
+        self.last_activity_time = time.monotonic()
```
FabienDanieau, line 135: Are all the time.monotonic related to this PR? Why the changes?

src/reachy_mini_conversation_app/tool_spaces.py
```diff
@@ -0,0 +1,414 @@
...
+        slug=validated_slug,
+        alias=alias,
+        mcp_url=mcp_url,
+        tags=sorted(space_info.tags or []),
+        tools=_build_installed_tool_space_tools(slug=validated_slug, alias=alias, remote_specs=remote_specs),
+        client=client,
+    )
+
+
+def resolve_public_tool_space_sync(slug: str) -> ResolvedInstalledToolSpace:
```
FabienDanieau, line 286: Didn't look into details but this loop management looks strange. Claude says:

> . `resolve_public_tool_space_sync` corrupts the thread's event-loop setting** — `tool_spaces.py:284-301`
> The finally block does `asyncio.set_event_loop(None)` whenever `previous_loop` is `None` (which is the normal case, since there's no *running* loop at startup). In the desktop-app path, `ReachyMiniConversationApp.run` at `main.py:338` explicitly calls `asyncio.set_event_loop(asyncio.new_event_loop())` — that gets wiped out by the first sync call into `_resolve_remote_tools`. `previous_loop = asyncio.get_running_loop()` does not see the *set* loop, only a running one, so the "restore previous loop" logic is incorrect.
> 
> Replace with `asyncio.run(resolve_public_tool_space(slug))` and stop touching the global loop policy. That's simpler and doesn't trample existing state.

src/reachy_mini_conversation_app/tools/core_tools.py
```diff
@@ -211,12 +298,71 @@ def _load_profile_tools() -> None:
...
+            logger.debug(
+                "Installed Space '%s' has no enabled tools in the active profile; skipping discovery.",
+                installed_space.slug,
+            )
+            continue
+
+        try:
+            resolved_space = resolve_public_tool_space_sync(installed_space.slug)
+        except Exception as exc:
+            raise RuntimeError(
```
FabienDanieau: Does it stop the program if a remotetool is not available? We can't consider that spaces will be always up and running

src/reachy_mini_conversation_app/tools/core_tools.py
```diff
@@ -89,6 +85,67 @@ async def __call__(self, deps: ToolDependencies, **kwargs: Any) -> Dict[str, Any
...
+        """Async tool execution entrypoint."""
+        ...
+
+
+ALL_TOOLS: Dict[str, ToolLike] = {}
+ALL_TOOL_SPECS: List[Dict[str, Any]] = []
+_TOOLS_INITIALIZED = False
+
+
+class RemoteMcpTool:
```
FabienDanieau: I'd prefer to have remotemcp tool to be a child of tool, and remove the toollike protocol. Looks like there is some duplication here. But it means to change the registration system. Not blocking for me, just saying.

src/reachy_mini_conversation_app/mcp_client.py
alozowski: Not quite, `__init__` is sync so we can't open an async session there, but actually we don't need to: the transport is streamable http, so there's no real persistent connection to share. httpx already pools TCP connections under the hood, so the socket is reused between calls. The only extra cost per call is `session.initialize()`, one HTTP POST – negligible compared to actual HF Space execution time

Please, correct me if I'm wrong

src/reachy_mini_conversation_app/base_realtime.py
alozowski: You're right, it's not related to this PR, it was a quick fix, because `asyncio.get_event_loop().time()` raises a `DeprecationWarning` in Python 3.10+ when called outside a running loop. It should have been a separate PR, so I can remove it from this work

src/reachy_mini_conversation_app/tool_spaces.py
alozowski: Claude is right, fixing 🤖

src/reachy_mini_conversation_app/tools/core_tools.py
alozowski: Yes it does right now, on my way to fix as well!

src/reachy_mini_conversation_app/tools/core_tools.py
alozowski: Agree, I will try to fix it minimally in this PR or I will do it in another one

alozowski: Done!

FabienDanieau (approved):
all good!

Issue #29, opened 2025-09-23
Dances description should be read from the dance library (and not hard coded) + add a forbidden list?


RemiFabre:
(simple fix)

The descriptions should be based on AVAILABLE_MOVES (cf https://github.com/pollen-robotics/reachy_mini_dances_library/blob/27171633a294622e7edb4f61cfc5dad951c367ec/src/reachy_mini_dances_library/collection/dance.py#L965)

and not like this:

    parameters_schema = {
        "type": "object",
        "properties": {
            "move": {
                "type": "string",
                "description": """Name of the move; use 'random' or omit for random. 
                                    Here is a list of the available moves: 
                                        simple_nod: A simple, continuous up-and-down nodding motion.
                                        head_tilt_roll: A continuous side-to-side head roll (ear to shoulder).
                                        side_to_side_sway: A smooth, side-to-side sway of the entire head.
                                        dizzy_spin: A circular 'dizzy' head motion combining roll and pitch.
                                        stumble_and_recover: A simulated stumble and recovery with multiple axis movements. Good vibes
                                        headbanger_combo: A strong head nod combined with a vertical bounce.
                                        interwoven_spirals: A complex spiral motion using three axes at different frequencies.
                                        sharp_side_tilt: A sharp, quick side-to-side tilt using a triangle waveform.
                                        side_peekaboo: A multi-stage peekaboo performance, hiding and peeking to each side.
                                        yeah_nod: An emphatic two-part yeah nod using transient motions.
                                        uh_huh_tilt: A combined roll-and-pitch uh-huh gesture of agreement.
                                        neck_recoil: A quick, transient backward recoil of the neck.
                                        chin_lead: A forward motion led by the chin, combining translation and pitch.
                                        groovy_sway_and_roll: A side-to-side sway combined with a corresponding roll for a groovy effect.
                                        chicken_peck: A sharp, forward, chicken-like pecking motion.
                                        side_glance_flick: A quick glance to the side that holds, then returns.
                                        polyrhythm_combo: A 3-beat sway and a 2-beat nod create a polyrhythmic feel.
                                        grid_snap: A robotic, grid-snapping motion using square waveforms.
                                        pendulum_swing: A simple, smooth pendulum-like swing using a roll motion.
                                        jackson_square: Traces a rectangle via a 5-point path, with sharp twitches on arrival at each checkpoint.
                """,
            },

Also, we should add a list of forbidden dance and emotions names.

[...]

Issue #313, opened 2026-04-22
Add dance move enum and fallback invalid moves to random


alozowski:
The dance tool should expose a strict valid move enum and normalise unknown moves to random.
Examples: the project tried to produce invalid `dance1` and `dance3` before retrying. In another run it lost the dance step after `dance(dance1)` failed.

alozowski:
Closing: #326 added the strict enum on the move argument plus the full move list in the tool description, which eliminates the invalid `dance1`/`dance3` calls this issue was about

PR #326, opened 2026-04-25
Fix/dance enum and fallback

yaseminozkut:
## Summary

- improves issue #313 
- replaced error return with `logger.info` and returning a random dance
- improved tool prompt and introduced `enum` list to restrict the choices (both for dance and emotion tool)

note: built on top of PR #186 since the changes are related (@RemiFabre closed the issue #29 but I added a comment back there, I think it is still needed for the llm description, what do you think?)

## Important notes
- `enum` as far as I found out, can only be fully strict with conversational openai models not realtime model. realtime openai model does not support `strict=True`. So adding `enum` will mostly for guiding the LLM.
- What I realized is there are emotions called dance1, dance2, and dance3 in the emotion library https://huggingface.co/datasets/pollen-robotics/reachy-mini-emotions-library/tree/main. So, since all tools are sent to llm together as a context, these names might be causing confusion and introducing a move as dance1, dance2, dance3. Maybe those emotions can be renamed or move to the dance library? their definition also involves "dance" word. 
- That's why I added `enum` to emotion tool as well and modified its prompt to be more strict. 
- However, since its not strict, its not exactly deterministic (just less hallucinating) but if similar hallucination happens in the future, now instead of error, it plays a random dance. 

Only changing dance tool still sometimes choose dance1, dance3 as either an emotion or a dance (still mixing up): 
before

After changing the emotion tool seems like fixed the results and now it only plays dance1, dance3 as emotion since its in its library: 
after

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [ ] Head wobble
</details>


RemiFabre:
Hi @yaseminozkut, thanks a lot for looking into this!

After re-discussing this with @alozowski , though, I think the requested change in #313 is not the direction we want to move forward with. It's fine for the LLM to request a random move, but the way that should work is by *not providing a name at all* (which is the current behavior). It's not okay that when the LLM makes a naming mistake, our system silently responds with a random move. I'd rather keep the current behavior where an error is sent back to the LLM, and then it's the LLM's responsibility to either retry or abandon. In general I'm fine with all sorts of non-repeatable behaviors from LLMs (I think it helps create the illusion of life) but the programatic tools we provide should be repeatable.

On the `dance1` / `dance2` / `dance3` naming overlap: it's fine if on a similar (or even the same) request, the LLM sometimes picks one of the emotion-dances and sometimes picks one of the dance-library dances. There's no harm in that. There is a technical separation between the two types of movements, dances are mathematical descriptions of motion (typically sin-based), and emotions are pre-recorded datasets of motion plus sound, but that differentiation isn't very important from the LLM's perspective. In my mind `dance1/2/3` are no different from any other entry in the dance library when it comes to the LLM choosing a dance. So I wouldn't stress about making that distinction very repeatable.

That said, I do agree we should strive to make the LLM commit as few naming mistakes as possible, so the `enum` change is good. And I think the behavior should be coherent between dances and emotions:

- Both use `enum`.
- Both return an error in case of an incorrect name.
- Both return a random move if no name is provided and this has to be clearly stated in the tool prompt.
- There is no `"random"` key: random is "empty input."

In general, I think we should try to keep changes as minimal as possible, to reduce the risk of breaking other behaviors that rely on this code.

Would you be OK implementing this in this PR? I'll gladly review it and merge it.

Thanks again. Also, the screenshots showing how you tested are very appreciated!

yaseminozkut:
Hi @RemiFabre , thanks for the feedback as well, I really appreciate it! I see what you mean. Yes of course, I would be happy to implement it while keeping both tools coherent as well.

yaseminozkut:
Hey @RemiFabre! I applied the changes, hope it's what you were looking for. Let me know if there is anything else. 

**One small note for the screenshots for reading it:** The chat UI prints in reverse order. Tool notifications appear before my question and the LLM's answer (only happens with Gemini), which is a bit confusing to follow. (I opened another issue+PR for that)

**Retry behavior:** The llm does seem like it can decide on its own. In another session, when it couldn't find the move, it decided not to retry. So, it looks like it can decide to repeat or abandon on its own like you said. Now the choices are more strict with `enum` and does not contain "random" in it. 

Screenshot 2026-04-27 at 5 35 37 PM
<br>

**Random move:** Now it is returning empty input instead of "random". The original behavior was mostly returning "random" because the prompt was "use 'random' or omit for random", so i only kept the omit part. In another session, it also picked a random dance move instead of passing "random" or empty input. Seems smart 😄
Screenshot 2026-04-27 at 6 28 22 PM
<br>

**Emotion tool:** To make emotion tool to omit for random and return empty input, I had to make `required` field empty as well in the schema. I tested out the emotion as well, 'enum' and slight change in the prompt reduced hallucination, though maybe more testing needed to confirm consistent behavior.

Screenshot 2026-04-27 at 6 36 45 PM

yaseminozkut:
Also, while testing, I realized when I want it to play an emotion, it is doing the motion related to the emotion but not playing the sound .wav files in the hugging face library. It is loaded in the code but never used in `EmotionQueueMove`. 

I can open an issue if its not actually intended this way, but since the robot speaks during emotion motion, playing the sound at the same time might conflict with the audio. Just wanted to ask before opening an issue.

RemiFabre:
> Also, while testing, I realized when I want it to play an emotion, it is doing the motion related to the emotion but not playing the sound .wav files in the hugging face library. It is loaded in the code but never used in `EmotionQueueMove`.
> 
> I can open an issue if its not actually intended this way, but since the robot speaks during emotion motion, playing the sound at the same time might conflict with the audio. Just wanted to ask before opening an issue.

Good question. We've tried different things and here is my convergence on this:
- In most cases, having the sound of the emotion + the sound of the voice together is not a good fit. They really sound like 2 different characters and there is nothing that guarantees that the emotion duration will match the speech duration (one of the motivation of [this btw](https://www.youtube.com/watch?v=MSl-0SfMU70))
- Sometimes, it would be great to have the sound of emotions. For example you say something mean to Mini and it goes "sad1" (without speaking) => great interaction


Ideally I would have wanted the LLM to be able to decide what to do. But in my tests, telling an LLM to answer without speaking is very unreliable (sometimes it speaks regardless). And forcing the mute programmatically is tricky for a bunch of reasons.

So my idea was to give the LLM a tool to toggle mute itself and give it another tool to toggle emotion sounds. I did this in this very old branch:
https://github.com/pollen-robotics/reachy_mini_conversation_app/tree/update_emotion_reader
which is an update on this pre-historic branch:
https://github.com/pollen-robotics/reachy_mini_conversation_app/tree/emotion_reader

The use case I had in mind was the "emotion reader" profile. You look at the robot and you make a face that expresses an emotion. The robot takes a picture, analyses it, and plays the closest emotion to it.
=> I never went through with this because gpt_realtime was too unreliable and wouldn't follow the script correctly. For example, the scoundrel gpt_realtime would always find a justification to unmute itself and speak, even though it was prompted not to.
But we now have a new [open source backend](https://github.com/pollen-robotics/reachy_mini_conversation_app/tree/codex/s2s-realtime-backend) that allows to choose the LLM (we haven't communicated on this yet, but it's going to be a big release), and now I believe it could work!

You're very welcome to pick up this work. For example, adding the 2 toggles discussed above as tools (so a cherry pick of "update_emotion_reader" on the current version of the software)

RemiFabre (approved):
Thanks for your work!

yaseminozkut:
@RemiFabre I remember seeing that video, it is actually so cool! I might try to extend it in the future. The new backend also sounds very exciting. 

Maybe if muting/unmuting toggle tool still creates problems even after the new backend, there can be a new tool that has a field in schema for speech where we don't pass it into `required = []` field. and based on the context, similar to the emotion and dance tools, it can either:
- omit speech for empty: if its empty it will use the emotion sound
- not empty: meaning LLM actually sends a LLM response that will be played while doing the emotion moves

But matching the speech duration with emotion duration seems tricky, if I can think of anything I'll ping you! 

Emotion reader idea is also very cool, like mirroring the user. Maybe if it is still performing unreliable after the backend change, it can be triggered with a voice activation/deactivation. But that way Reachy will only be in emotion reader/mirroring state and won't be in the general conversation mode. It can be a workaround as a last option.

I'd be very happy to explore and pick up the emotion reading feature or in general emotion sounds vs. speech picking. See what I can do after the new backend update.

PR #331, opened 2026-04-27
Set audio board startup defaults

andimarafioti:
## Summary

- Keep the conversation app's tuned audio-board startup parameter set.
- Apply that parameter set through the Reachy Mini SDK audio config API added in pollen-robotics/reachy_mini#1057 instead of directly opening or decoding the ReSpeaker control interface here.
- Keep startup application in the app and console flows, with simulation mode still skipped.

## Validation

- `uv run pytest tests/audio/test_startup_config.py`
- `uv run ruff check`
- `uv run ruff format --check src/reachy_mini_conversation_app/audio/startup_config.py tests/audio/test_startup_config.py`
- `uv run mypy src/reachy_mini_conversation_app/audio/startup_config.py`
- `git diff --check`

### DIFF
--- a/src/reachy_mini_conversation_app/audio/startup_config.py
+++ b/src/reachy_mini_conversation_app/audio/startup_config.py
@@ -1,0 +1,173 @@
+"""Startup configuration for the Reachy Mini audio processor."""
+
+from __future__ import annotations
+import time
+import struct
+import logging
+from typing import Protocol
+from collections.abc import Callable, Sequence
+
+from reachy_mini.media.audio_control_utils import PARAMETERS, init_respeaker_usb
+
+
+AudioControlValue = float | int
+AudioStartupParameter = tuple[str, tuple[AudioControlValue, ...]]
+WRITE_SETTLE_SECONDS = 0.1
+VERIFY_TOLERANCE = 1e-3
+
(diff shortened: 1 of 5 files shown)


src/reachy_mini_conversation_app/audio/startup_config.py
```diff
@@ -0,0 +1,173 @@
...
+    ("PP_MIN_NS", (0.8,)),
+    ("PP_MIN_NN", (0.8,)),
+    ("PP_GAMMA_E", (0.5,)),
+    ("PP_GAMMA_ETAIL", (0.5,)),
+    ("PP_NLATTENONOFF", (0,)),
+    ("PP_MGSCALE", (4.0, 1.0, 1.0)),
+)
+
+
+class ReSpeakerControl(Protocol):
```
FabienDanieau: Would it be possible to use this class to interact with the respeaker? https://github.com/pollen-robotics/reachy_mini/blob/2be8d323ffd04ee8ceb62b096e2738df6697be7e/src/reachy_mini/media/audio_control_utils.py#L174

andimarafioti: That's a great point, I missed that, thank you!

FabienDanieau (approved):
Perfect!

PR #336, opened 2026-04-29
335 head wobbler is now handled on the core side

### LINKED ISSUE #335: Head wobbler is now handled on the core side

RemiFabre:
# Summary

Drop the in-app head wobbler, use the core daemon's instead.

The `reachy_mini` SDK now ships its own audio-reactive head wobbler that runs inside the daemon's media pipeline. Since we already used`push_audio_sample`, this PR is very simple. The local DSP, threading, scheduling, and offset-application code we maintained in the app are no longer needed.

Roughly 970 deletions vs 11 insertions.

## What changes

- **`main.py`**: replace the local `HeadWobbler` lifecycle with `robot.enable_wobbling()` / `robot.disable_wobbling()`.
- **Backends (`gemini_live.py`, `openai_realtime.py`)**: drop the `if deps.head_wobbler is not None:` blocks that called `.reset()`, `.request_reset_after_current_audio()`, and `.feed()`. The daemon's wobbler reacts to the audio stream itself, so explicit reset hooks are unnecessary.
- **`console.py`**: drop the `head_wobbler.feed_pcm` call and the `_estimate_pending_playback_seconds` helper that only existed to compute its `start_delay_s`.
- **Delete `src/reachy_mini_conversation_app/audio/`**: `head_wobbler.py`, `speech_tapper.py`, and the package itself.
- **`MovementManager`**: drop `set_speech_offsets` and the supporting state (`_pending_speech_offsets`, `_speech_offsets_lock`, `_speech_offsets_dirty`, `state.speech_offsets`, the speech term in `_apply_pending_offsets` and `_get_secondary_pose`). Speech offsets are now composed server-side via `SetSpeechOffsetsCmd`; the local composition path is gone.
- **`ToolDependencies`**: drop the `head_wobbler` field.
- **Tests**: drop `tests/audio/test_head_wobbler.py`, the two head-wobbler-only tests in `tests/test_openai_realtime.py`, the playback-delay test in `tests/test_console.py`, and update `tests/test_gemini_live.py`'s remaining wobbler-aware test to drop the wobbler mock while keeping its transcript and listening-state coverage.

## Mathematical equivalence

Side note: the implementation on the core side is almost identical to the one we had here. The only change is an improvement, basically:
final_new = compose_world_offset( compose_world_offset(primary, face), speech )
final_old = compose_world_offset( primary, single_matrix(speech + face) )

-> The impact should not be noticeable but it's still an improvement.

## Behaviour change to be aware of

In `--gradio` mode, audio is rendered in the browser and never reaches the daemon's media pipeline, so the head no longer wobbles in that path. Production (non-gradio) paths use `push_audio_sample`, which routes audio through the daemon's pipeline and the wobbler works.

## Test plan

- [x] `pytest tests/test_console.py tests/test_openai_realtime.py` passes (24 / 24).
- [x] Manual test on a lite robot in non-gradio mode

## Notes
linked PR from the SDK side: https://github.com/pollen-robotics/reachy_mini/pull/1001


src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -705,8 +705,6 @@ async def _run_realtime_session(self) -> None:
                         self._turn_first_audio_at = None
                         if hasattr(self, "_clear_queue") and callable(self._clear_queue):
                             self._clear_queue()
-                        if self.deps.head_wobbler is not None:
-                            self.deps.head_wobbler.reset()
                         self.deps.movement_manager.set_listening(True)
```
alozowski, line 732: I see that the `wobbler.reset()` call was removed above, was this removed on purpose? Just want to note for myself

src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -810,8 +806,6 @@ async def _run_realtime_session(self) -> None:
                     if event.type == "response.output_audio.delta":
```
alozowski, line 818: In Gradio mode, we used to feed each audio chunk to the head wobbler here, but with the current changes, the head won't wobble during the robot's speech in Gradio mode. Let's restore the Gradio mode functionality, please (though it will be removed soon)

src/reachy_mini_conversation_app/main.py
```diff
@@ -272,7 +268,11 @@ def run(
 
     # Each async service → its own thread/loop
     movement_manager.start()
-    head_wobbler.start()
+    # Audio-reactive head motion is driven by the daemon's wobbler. The
```
alozowski: This comment only describes what happens in local (headless) mode, where audio goes through `push_audio_sample`. In Gradio mode the audio takes a different path and the daemon wobbler doesn't see it. Let's narrow this comment, it will be easier to update it in the future

src/reachy_mini_conversation_app/moves.py
```diff
@@ -566,16 +527,13 @@ def _get_primary_pose(self, current_time: float) -> FullBodyPose:
...
-            self.state.speech_offsets[0] + self.state.face_tracking_offsets[0],
-            self.state.speech_offsets[1] + self.state.face_tracking_offsets[1],
-            self.state.speech_offsets[2] + self.state.face_tracking_offsets[2],
-            self.state.speech_offsets[3] + self.state.face_tracking_offsets[3],
-            self.state.speech_offsets[4] + self.state.face_tracking_offsets[4],
-            self.state.speech_offsets[5] + self.state.face_tracking_offsets[5],
-        )
+        """Get the secondary full body pose from face tracking offsets.
+
+        Speech-induced offsets are now produced by the daemon's wobbler and
```
alozowski: This sentence describes a past change rather than what the function does. It will become confusing over time. Could we remove it? The first line of the docstring is already clear enough on its own

alozowski (changes requested):
It's nice to see how cleaner the project is with this work! The main thing before merging: Gradio wobbling is silently broken – let's fix it minimally, even despite we are going to remove it very soon

src/reachy_mini_conversation_app/audio/head_wobbler.py
```diff
@@ -1,211 +0,0 @@
-"""Moves head given audio samples."""
-
-import time
-import queue
-import base64
-import logging
-import threading
```
andimarafioti, line 7: RIP

src/reachy_mini_conversation_app/base_realtime.py
FabienDanieau: What do you mean?

FabienDanieau:
> It's nice to see how cleaner the project is with this work! The main thing before merging: Gradio wobbling is silently broken – let's fix it minimally, even despite we are going to remove it very soon

My solution is to also push audio sample to a mute reachy mini in the case of gradio. So we still use the core wobbler

src/reachy_mini_conversation_app/base_realtime.py
FabienDanieau: There is now a `_tap_audio_for_daemon_wobbler` function that foward audio samples to reachy mini so the wobbling works (and the audio output of the robot is muted)

src/reachy_mini_conversation_app/base_realtime.py
alozowski: I missed the comment 😱
I mean are we sure to remove the `head_wobbler.reset()` here?

FabienDanieau: yes, the wobbler is reset on the core side when the audio is asked to be flushed.

alozowski (approved):
Nice!

PR #360, opened 2026-05-20
295 add memory in s2s realtime backend branch

### LINKED ISSUE #295: Add memory in s2s-realtime-backend branch

RemiFabre:
Please read the design document here. I added a few verbatim copies of my memory files to have real examples:
https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/4fccb367234342c0915d53ebde10144e7b12fafa/docs/memory-system-design.md


RemiFabre:
@andimarafioti One problem I see is that currently I use an API key to call the LLM that does the "dream" phase. This was easy when every conv app had an API key, but this won't translate to the HF backend out of the box without changes to the speech2speech repo. Thoughts?

andimarafioti:
We could move this into speech-to-speech and have it call the LLM directly in the same way that we do compaction 💡

RemiFabre:
After discussing this with Andi, we've converged on implementing the dreaming process server side (in the speech_to_speech repo).
I think this branch should still be tested as I've made a bunch of design choices that need to be tested by more than 1 person.

PR #362, opened 2026-05-22
Make idle tools locally random

andimarafioti:
## Summary

Idle behavior is now selected locally by the app instead of being prompted through a realtime model turn. The local idle policy keeps `idle_do_nothing` as the common case and occasionally selects available movement tools such as `dance`, `play_emotion`, or `move_head`.

## Details

- Added a small shared `idle_policy.py` helper for choosing the local idle tool call.
- Updated OpenAI-compatible realtime and Gemini Live idle handling to start a local background tool instead of sending idle text/model tool-choice requests.
- Kept idle tool results out of the model conversation so local idle actions do not trigger speech follow-ups.
- Kept `idle_do_nothing`, profile tool lists, prompts, docs, and normal LLM-facing tool specs unchanged.

## Validation

- `uv run ruff check src/reachy_mini_conversation_app/idle_policy.py src/reachy_mini_conversation_app/base_realtime.py src/reachy_mini_conversation_app/gemini_live.py tests/test_openai_realtime.py tests/test_gemini_live.py`
- `uv run pytest tests/test_openai_realtime.py tests/test_gemini_live.py` (`32 passed`)


src/reachy_mini_conversation_app/idle_policy.py
```diff
@@ -0,0 +1,30 @@
...
+import random
+from typing import Any, Final, Iterable
+
+
+_IDLE_TOOL_WEIGHTS: Final[tuple[tuple[str, float], ...]] = (
+    ("idle_do_nothing", 0.60),
+    ("dance", 0.16),
+    ("play_emotion", 0.16),
+    ("move_head", 0.08),
+)
```
alozowski: These strings are an implicit contract with the tool classes – if a tool is renamed it silently falls out of idle selection, and when someone adds a new tool there's no obvious path that leads them here. The `if` branches for argument generation are also split away from the weight declaration. Could we make this explicit and co-located? For example, `idle_weight` and `idle_args` on the `Tool` base class so each tool declares its own eligibility and arguments, and `choose_idle_tool_call` builds the candidates from the registry instead of a hardcoded list. Also worth adding a module docstring to `idle_policy.py` documenting the effective distribution – once weights are scattered across tool files it's no longer readable in one place (so we'll have to commit to keeping that docstring up to date, but it will be much clearer)

src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -1016,23 +1016,48 @@ async def _build_realtime_client(self) -> AsyncOpenAI:
         """Build the realtime SDK client for this backend."""
 
     async def send_idle_signal(self, idle_duration: float) -> None:
```
alozowski, line 1015: The body of this method is now duplicated in `GeminiLiveHandler.send_idle_signal`. Could we extract the shared logic into a helper in `idle_policy.py` that takes `tool_manager`, `output_queue`, `available_tool_names`, and `idle_duration`? Both handlers would then just resolve their tool specs and delegate

src/reachy_mini_conversation_app/gemini_live.py
```diff
@@ -725,19 +734,48 @@ def format_timestamp(self) -> str:
         return f"[{dt.strftime('%Y-%m-%d %H:%M:%S')} | +{elapsed_seconds:.1f}s]"
 
     async def send_idle_signal(self, idle_duration: float) -> None:
```
alozowski, line 731: This is an almost exact copy of `BaseRealtimeHandler.send_idle_signal` – any future change to idle dispatch has to land in two places. See the comment there about extracting a shared helper

src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -1016,23 +1016,48 @@ async def _build_realtime_client(self) -> AsyncOpenAI:
...
-            response=RealtimeResponseCreateParamsParam(
-                instructions="You MUST respond with function calls only - no speech or text. Choose appropriate actions for idle behavior. Use idle_do_nothing only if you intentionally want no movement or sound during this idle turn.",
-                tool_choice="required",
+        await self.output_queue.put(
+            AdditionalOutputs(
+                {
+                    "role": "assistant",
+                    "content": (
+                        f"🛠️ Idle tool {tool_name} with args {args_json_str}. "
+                        f"The tool is now running. Tool ID: {bg_tool.tool_id}"
```
alozowski: The Gemini version of this message doesn't have "The tool is now running." – looks like it got dropped during copy. Could you please align the two?

alozowski (changes requested):
I like this idea! I left a couple of code comments and one architectural thought

RemiFabre (approved):
This is a nice change, I shouldn't have designed this way in the first place (too expensive for little extra value). One regression this has though is that we lose the per profile behaviors. This means that the "noir detective" profile will now randomly dance on ~16% of idle ticks, breaking character.

We could make it so profiles can declare their weighted list of plausible idle moves. But I'm not sure the added complexity is worth it. I'd merge this as as, and then iterate later if needed.


Btw I'm not 100% sure these idle behaviors are worth it? If you want to straight up remove them we can discuss/brainstorm it

RemiFabre:
btw I think the class attribute self.is_idle_tool_call is now orphaned. Maybe delete it from these 2 places:
https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5894dde185e30a9e5f70a980f2c7f85b376bb914/src/reachy_mini_conversation_app/base_realtime.py#L137

https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5894dde185e30a9e5f70a980f2c7f85b376bb914/src/reachy_mini_conversation_app/gemini_live.py#L169

src/reachy_mini_conversation_app/idle_policy.py
andimarafioti: I agree with not passing hardcoded names, but I would not have the tools have their own weighting. Instead, I would use the class type here and still have hardcoded values for them. Also, following remi's comment below, those values could be changed per profile

src/reachy_mini_conversation_app/base_realtime.py
andimarafioti: looks like a good simplification to tackle, codex, look at it

src/reachy_mini_conversation_app/gemini_live.py
andimarafioti: codex, address this

src/reachy_mini_conversation_app/base_realtime.py
andimarafioti: Codex!

alozowski: even the soulless machine doesn't work to work late evening 😂

src/reachy_mini_conversation_app/idle_policy.py
andimarafioti: Codex here: addressed in `8db9ba2`. `choose_idle_tool_call` now uses hardcoded class-type weighted candidates instead of matching raw tool-name strings, keeps each candidate args factory next to its weight, and filters those candidates against the active registered tools. I also added the module docstring with the effective default distribution and focused idle-policy tests.

src/reachy_mini_conversation_app/base_realtime.py
andimarafioti: Codex here: addressed in `8db9ba2`. The shared idle dispatch path now lives in `idle_policy.start_idle_tool_call`; the OpenAI-compatible handler only resolves active tool names and delegates to it. I included `deps` as an extra helper argument because `ToolCallRoutine` still needs the current session dependencies.

src/reachy_mini_conversation_app/gemini_live.py
andimarafioti: Codex here: addressed in `8db9ba2`. `GeminiLiveHandler.send_idle_signal` now delegates to the same `idle_policy.start_idle_tool_call` helper as `BaseRealtimeHandler`, so the duplicated idle start-tool body is gone.

src/reachy_mini_conversation_app/base_realtime.py
andimarafioti: Codex here: addressed in `8db9ba2`. The idle console message is emitted by the shared helper now, so OpenAI-compatible realtime and Gemini both include `The tool is now running.`; the focused OpenAI/Gemini idle tests assert the aligned message.

src/reachy_mini_conversation_app/idle_policy.py
alozowski: Let's go with class types then, though it worth adding a module docstring too for readability

andimarafioti:
Codex here, @RemiFabre: I addressed your concrete orphaned-flag comment in `52e1d3a`.

- Removed `self.is_idle_tool_call` from both `BaseRealtimeHandler` and `GeminiLiveHandler`.
- Model-originated tool calls now explicitly start background tools with `is_idle_tool_call=False`.
- Locally selected idle tools are still marked as idle at the background-tool level by `idle_policy.start_idle_tool_call`, so idle tool results remain kept out of the model conversation.

On the per-profile behavior point: I did not add per-profile idle weights in this pass. The policy is still filtered through the active tool specs and renormalized across available idle candidates; profile-specific weighted behavior can stay as a follow-up since your review said merge as-is and iterate later if needed.

src/reachy_mini_conversation_app/idle_policy.py
```diff
@@ -0,0 +1,141 @@
...
+from reachy_mini_conversation_app.tools.background_tool_manager import (
+    BackgroundTool,
+    ToolCallRoutine,
+    BackgroundToolManager,
+)
+
+
+logger = logging.getLogger(__name__)
+
+_IDLE_MOVE_HEAD_DIRECTIONS: Final[tuple[str, ...]] = ("left", "right", "up", "down", "front")
```
alozowski: can we use MoveHead deltas here?

alozowski (approved):
just one question left but approved!

PR #374, opened 2026-05-27
Keep settings UI available on backend failure

andimarafioti:
## Summary

- Keep the headless settings UI available when the active realtime backend fails to connect or exits.
- Expose backend connection state and the last startup error through `/status`.
- Add a visible settings-page banner for connecting, disconnected, waiting-for-config, and restart-required states.
- Add tests covering backend failure status reporting and non-crashing startup behavior.

## Root Cause

`LocalStream` launched the realtime backend task in the same gathered task set as record/play loops. If the backend startup raised, the exception unwound the runner and the settings UI was no longer useful for changing away from a broken local Hugging Face target.

## Impact

When Hugging Face local mode points at a server that is not running, users can still open settings, see that the backend is disconnected, and switch to the built-in server or another backend without first starting the local server.

## Validation

- `uv run pytest`
- `uv run ruff check`
- `uv run mypy`
- `node --check src/reachy_mini_conversation_app/static/main.js`


src/reachy_mini_conversation_app/static/style.css
```diff
@@ -165,6 +165,41 @@ body {
...
+  line-height: 1.5;
+}
+
+.connection-alert.error {
+  border-color: rgba(255, 92, 112, 0.42);
+  background: rgba(255, 92, 112, 0.1);
+  color: var(--error);
+}
+
+.connection-alert.ok {
```
alozowski: `connectionAlert.className` is only ever set to `connection-alert hidden`, `connection-alert`, or `connection-alert error`, but `.ok` is never applied. Was this intended for a future version, or can we safely remove it?

src/reachy_mini_conversation_app/console.py
```diff
@@ -514,6 +552,47 @@ async def _validate_key(payload: ApiKeyPayload) -> JSONResponse:
...
+                    "%s backend failed to start: %s. Settings UI remains available; retrying in %.1f seconds.",
+                    active_backend,
+                    e,
+                    self._backend_retry_delay,
+                    exc_info=logger.isEnabledFor(logging.DEBUG),
+                )
+            else:
+                if self._stop_event.is_set():
+                    return
+                self._set_backend_connection_state("disconnected", "Realtime session ended.")
```
alozowski: When `start_up()` returns without raising, `"Realtime session ended."` gets stored as `backend_error` and surfaces as "Last error:" in the settings page. A normal session end isn't a failure, so dropping the second argument fixes it:
```
self._set_backend_connection_state("disconnected")
```

src/reachy_mini_conversation_app/console.py
```diff
@@ -164,6 +168,38 @@ def _active_backend(self) -> str:
...
+        if message:
+            return f"{type(error).__name__}: {message}"
+        return type(error).__name__
+
+    def _set_backend_connection_state(self, state: str, error: BaseException | str | None = None) -> None:
+        """Update backend connection status exposed through the settings UI."""
+        self._backend_connection_state = state
+        if error is not None:
+            self._backend_error = self._format_backend_error(error)
+        elif state in {"not_started", "waiting_for_config", "connecting", "connected", "restart_required"}:
```
alozowski: The set lists every state that should clear the error, but "disconnected" is the only one that shouldn't. Inverting the condition is cleaner and won't silently break when a new state is added:
```
elif state != "disconnected":
    self._backend_error = None
```
Also, "connected" in the current set is never passed to this method, that state is computed in `_backend_connection_status()`, not set here.

src/reachy_mini_conversation_app/console.py
```diff
@@ -544,20 +623,30 @@ def launch(self) -> None:
...
                     "%s not found. Set it in the app .env before starting the Hugging Face backend.", requirement_name
                 )
                 return
-            else:
-                logger.warning("%s not found. Open the app settings page to enter it.", requirement_name)
+            logger.warning("%s not found. Open the app settings page to configure it.", requirement_name)
             # Poll until the key becomes available (set via the settings UI)
             try:
-                while not self._has_required_key(active_backend):
+                while not self._stop_event.is_set() and not self._has_required_key(active_backend):
```
alozowski, line 635: `waiting_for_config` is already set before this loop at L626, so the else branch just repeats the same value every 200 ms. The loop only needs to react when the backend choice actually changes:
```
while not self._stop_event.is_set() and not self._has_required_key(active_backend):
    if get_backend_choice() != active_backend:
        self._set_backend_connection_state("restart_required")
    time.sleep(0.2)
```
The reset at L649 already handles the exit

src/reachy_mini_conversation_app/static/main.js
```diff
@@ -462,9 +540,12 @@ async function init() {
   show(configuredPanel, false);
   show(personalityPanel, false);
 
-  const st = (await waitForStatus()) || {
+  let st = (await waitForStatus()) || {
     active_backend: DEFAULT_BACKEND,
     backend_provider: DEFAULT_BACKEND,
+    backend_connected: false,
+    backend_connection_state: "unknown",
+    backend_error: "",
```
alozowski: `backend_error` is `null` in the API response, should this be null too?

tests/test_console.py
```diff
@@ -429,6 +429,70 @@ def test_status_reports_direct_hf_ws_url_as_ready(
...
+    stream = LocalStream(handler, robot, instance_path=str(tmp_path))
+    stream._backend_retry_delay = 0
+
+    async def fail_and_stop() -> None:
+        stream._stop_event.set()
+        raise RuntimeError("local server unavailable")
+
+    handler.start_up = AsyncMock(side_effect=fail_and_stop)
+
+    await asyncio.wait_for(stream._run_handler_startup_loop(), timeout=1.0)
```
alozowski: This calls `_run_handler_startup_loop` directly, since it's the main loop rather than a small helper, it's probably fine, but it does couple the test to an internal name. The same assertions could go through `GET /status`, same pattern as `test_status_reports_backend_connection_failure` above

alozowski (changes requested):
The approach is solid! I left a few small comments, mostly cleanup

src/reachy_mini_conversation_app/static/style.css
andimarafioti: Removed `.connection-alert.ok` in `534d8ba` since the alert only renders the default warning style or `.error`.

src/reachy_mini_conversation_app/console.py
andimarafioti: Changed this in `534d8ba` to clear the error for every non-`disconnected` state with `elif state != "disconnected"`, so future non-error states clear safely too.

tests/test_console.py
andimarafioti: Updated the test in `534d8ba` to drive `stream.launch()` and assert the result through `GET /status`; record/play loops are mocked, and the default asyncio loop is restored afterward so the test does not leak loop state.

src/reachy_mini_conversation_app/console.py
andimarafioti: Updated this in `534d8ba`: normal `start_up()` return now calls `_set_backend_connection_state("disconnected")` without setting `backend_error`, and the log is `info` rather than `warning`.

src/reachy_mini_conversation_app/console.py
andimarafioti: Updated in `534d8ba`: the loop no longer rewrites `waiting_for_config` every 200 ms and now only reacts when the backend choice changes.

src/reachy_mini_conversation_app/static/main.js
andimarafioti: Changed the fallback to `backend_error: null` in `534d8ba` to match the API response shape.

alozowski (approved):
Let's merge!

PR #377, opened 2026-05-29
Centralize headless backend restarts

andimarafioti:
## Summary
- move headless backend rebuild/retry ownership into LocalStream using a handler factory
- let backend config, personality, and voice changes request an in-process backend restart instead of requiring app restart or backend-owned reconnects
- update the settings UI to show reconnecting status without reloading, and add LocalStream-level regression coverage

## Verification
- uv run ruff format --check
- uv run ruff check
- uv run mypy
- node --check src/reachy_mini_conversation_app/static/main.js
- uv run pytest -q

### DIFF
--- a/src/reachy_mini_conversation_app/console.py
+++ b/src/reachy_mini_conversation_app/console.py
@@ -120,7 +120,6 @@ class LocalStream:
         - ``instance_path``: directory where per-instance ``.env`` should be stored.
         - ``handler_factory``: builds a fresh handler for the currently selected backend.
         """
-        self.handler = handler
         self._robot = robot
         self._stop_event = asyncio.Event()
         self._restart_requested = asyncio.Event()
@@ -223,7 +222,6 @@ class LocalStream:
 
     async def request_backend_restart(self, reason: str) -> None:
         """Ask the startup loop to rebuild the backend and stop the current handler."""
-        logger.info("Backend restart requested: %s", reason)
         self._set_backend_connection_state("connecting")
         self._restart_requested.set()
         await self._shutdown_active_handler()
@@ -449,14 +447,14 @@ class LocalStream:
             except BaseException:
                 set_custom_profile(previous_profile)
                 raise
-            await self.request_backend_restart("personality_changed")
-            return "Applied personality and restarting backend."
         except Exception as e:
             logger.error("Error applying personality '%s': %s", profile, e)
             return f"Failed to apply personality: {e}"
         except BaseException as e:
             logger.error("Failed to resolve personality content: %s", e)
             return f"Failed to apply personality: {e}"
+        await self.request_backend_restart("personality_changed")
+        return "Applied personality and restarting backend."
 
     async def get_available_voices(self) -> list[str]:
         """Return voices available for the currently selected backend."""
--- a/src/reachy_mini_conversation_app/console.py
+++ b/src/reachy_mini_conversation_app/console.py
@@ -15,7 +15,7 @@ import sys
 import time
 import asyncio
 import logging
-from typing import List, Optional
+from typing import List, Callable, Optional
 from pathlib import Path
 
 from fastrtc import AdditionalOutputs, audio_to_float32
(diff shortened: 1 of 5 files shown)


src/reachy_mini_conversation_app/console.py
```diff
@@ -108,18 +111,22 @@ def __init__(
...
+        handler_factory: HandlerFactory | None = None,
+        startup_voice: Optional[str] = None,
     ):
         """Initialize the stream with a realtime handler and pipelines.
 
         - ``settings_app``: the Reachy Mini Apps FastAPI to attach settings endpoints.
         - ``instance_path``: directory where per-instance ``.env`` should be stored.
+        - ``handler_factory``: builds a fresh handler for the currently selected backend.
         """
         self.handler = handler
```
alozowski: This assignment is now redundant, because `_install_handler(handler)` at the bottom of `__init__` sets `self.handler` again. Can we drop this line?

src/reachy_mini_conversation_app/console.py
```diff
@@ -170,7 +183,59 @@ def _active_backend(self) -> str:
...
+        self._set_backend_connection_state("connecting")
+        loop = self._asyncio_loop
+        if loop is not None and loop.is_running():
+            asyncio.run_coroutine_threadsafe(self.request_backend_restart(reason), loop)
+            return
+        self._restart_requested.set()
+
+    async def request_backend_restart(self, reason: str) -> None:
+        """Ask the startup loop to rebuild the backend and stop the current handler."""
+        logger.info("Backend restart requested: %s", reason)
```
alozowski: This line is a duplicate: `_mark_restart_requested` already logs the same message before calling this method, and `apply_personality`/`change_voice` log their own context before awaiting it. Can you remove this logger call?

src/reachy_mini_conversation_app/console.py
```diff
@@ -370,6 +435,60 @@ def _read_persisted_personality(self) -> Optional[str]:
...
+                get_session_voice(default=get_default_voice_for_backend(get_backend_choice()))
+            except BaseException:
+                set_custom_profile(previous_profile)
+                raise
+            await self.request_backend_restart("personality_changed")
+            return "Applied personality and restarting backend."
+        except Exception as e:
+            logger.error("Error applying personality '%s': %s", profile, e)
+            return f"Failed to apply personality: {e}"
+        except BaseException as e:
```
alozowski, line 453: This except `BaseException` also wraps await `request_backend_restart`, so if the task gets cancelled, `asyncio.CancelledError` is caught here and silently turned into a return string – that breaks task cancellation. Could you move `await request_backend_restart` and its return outside the try block so this guard only covers the prompt loading?

alozowski (changes requested):
Nice work! Just a few minor things to polish

src/reachy_mini_conversation_app/console.py
andimarafioti: Codex here: addressed in `dc7e8f8`. I removed the redundant `self.handler = handler` assignment from `LocalStream.__init__`; `_install_handler(handler)` remains responsible for setting and wiring the active handler.

src/reachy_mini_conversation_app/console.py
andimarafioti: Codex here: addressed in `dc7e8f8`. I removed the duplicate `logger.info` from `request_backend_restart`, leaving the contextual logging in `_mark_restart_requested`, `apply_personality`, and `change_voice`.

src/reachy_mini_conversation_app/console.py
andimarafioti: Codex here: addressed in `dc7e8f8`. I moved `await request_backend_restart("personality_changed")` and the success return outside the prompt-loading try block, so `asyncio.CancelledError` from the restart path propagates instead of becoming a status string. I also added `test_apply_personality_propagates_restart_cancellation` to cover this.

alozowski (approved):
let's merge!

PR #386, opened 2026-06-02
Flush WebRTC playback on barge-in via clear_player()

### LINKED ISSUE #376: WebRTC audio is not actually flushed on user interruption
## Summary

When the conversation app runs against a wireless Reachy Mini over WebRTC, user barge-in/interruption reaches the app, but previously queued assistant audio can continue playing. This makes it sound like the assistant keeps talking over the user until the next turn starts, even though the backend has already detected the interruption.

This was reproduced with both the local speech-to-speech realtime backend and OpenAI Realtime, so the issue does not appear to be backend-specific.

## Evidence

Recent OpenAI Realtime logs show the existing interruption signal path firing:

```text
2026-05-28 15:10:48,909 INFO reachy_mini_conversation_app.console:631 | User intervention: flushing player queue
2026-05-28 15:10:55,270 INFO reachy_mini_conversation_app.console:631 | User intervention: flushing player queue
2026-05-28 15:10:55,291 INFO reachy_mini_conversation_app.console:674 | role=user content=The last time.
2026-05-28 15:10:56,149 INFO reachy_mini_conversation_app.console:674 | role=user content=Life
```

So `input_audio_buffer.speech_started` is being received and the app calls its queue-clearing hook before the user transcript is completed.

The conversation app has the expected wiring:

- `base_realtime.py` handles `input_audio_buffer.speech_started` by calling `_clear_queue()`.
- `console.py` assigns `self.handler._clear_queue = self.clear_audio_queue`.
- `clear_audio_queue()` logs `User intervention: flushing player queue` and, for `MediaBackend.WEBRTC`, calls `audio.clear_output_buffer()`.

However, in the installed Reachy Mini WebRTC media backend, `clear_output_buffer()` is currently a no-op:

```python
def clear_output_buffer(self) -> None:
    """No-op (WebRTC send chain does not buffer significantly)."""
    pass
```

That means the app-level flush hook runs, but the wireless/WebRTC audio path does not actually clear queued playback.

## Expected Behavior

When the user starts speaking during assistant audio playback:

1. The realtime backend emits/receives the existing speech-start interruption signal.
2. The app clears queued assistant audio.
3. The robot stops speaking promptly, without playing already buffered assistant audio.

## Actual Behavior

The app logs that it is flushing the player queue, but on the wireless/WebRTC robot the old assistant audio may keep playing. The assistant then answers the interrupted/new user turn, which makes it appear as if it heard the user while still talking over them.

## Likely Cause

The WebRTC media implementation does not implement an effective output buffer flush. Audio pushed into the GStreamer/WebRTC send path may already be queued downstream in `appsrc`, encoder/payloader/webrtcbin, UDP/WebRTC buffers, or the robot receiver.

There is also a smaller app-side concern: `clear_audio_queue()` currently replaces `handler.output_queue` with a new `asyncio.Queue()`. If the playback loop is already awaiting the old queue, replacing the object can be brittle. Draining the existing queue in place would be safer.

## Suggested Fix

Implement a real `clear_output_buffer()` for the WebRTC audio backend. At minimum it should flush the GStreamer send chain and reset audio timestamps, similar to the local GStreamer player flush path.

Potential approach:

```python
def clear_output_buffer(self) -> None:
    if self._appsrc is None:
        return

    self._appsrc_pts = 0
    self._pipeline_record.set_state(Gst.State.PAUSED)
    self._appsrc.send_event(Gst.Event.new_flush_start())
    self._appsrc.send_event(Gst.Event.new_flush_stop(reset_time=True))
    self._pipeline_record.set_state(Gst.State.PLAYING)
    self.logger.info("Cleared WebRTC output buffer")
```

This may need to flush a broader part of the WebRTC send chain depending on where the audio is actually buffered.

It may also help to make the WebRTC send path explicitly low-latency/leaky, so several seconds of assistant audio cannot accumulate downstream.

In the conversation app, consider changing queue clearing from replacing the queue object to draining it in place:

```python
while True:
    try:
        self.handler.output_queue.get_nowait()
    except asyncio.QueueEmpty:
        break
```

## Notes

This is not fixed by adding a new realtime protocol event: the existing OpenAI-style `input_audio_buffer.speech_started` path already reaches the app. The missing piece is making the WebRTC playback flush actually clear buffered output audio.

FabienDanieau:
## Summary

Fixes #376. On the WebRTC backend, user barge-in reached the app but the robot kept playing already-buffered assistant audio.

Two causes:
1. `clear_audio_queue()` called the SDK's `clear_output_buffer()` for WebRTC, which is now a deprecated no-op. The real flush is `clear_player()`.
2. It replaced `handler.output_queue` with a fresh `asyncio.Queue()`; a blocked `emit()`/`wait_for_item()` keeps draining the *old* queue and can replay stale audio.

## Changes (`console.py`)

- Call `audio.clear_player()` uniformly (now a first-class flush on both the local GStreamer and WebRTC backends — the WebRTC one also tells the daemon to drop audio queued for the speaker), with a `clear_output_buffer()` fallback for older SDKs.
- Drain `handler.output_queue` **in place** (`get_nowait()` loop) instead of replacing it.
- Remove the now-unused `MediaBackend` import and `LOCAL_PLAYER_BACKEND` constant.

## Dependency

The effective WebRTC flush requires a `reachy-mini` that ships the `clear_player()` WebRTC flush (pollen-robotics/reachy_mini#1186). The `hasattr` guard keeps this safe on older SDKs; bump the `reachy-mini` pin once that release is out.

## Verification

- `ruff check` + `mypy` (strict) clean on `console.py`.
- Barge-in on a wireless robot: robot stops promptly; SDK logs `Cleared player queue (WebRTC send chain flushed)`, daemon logs `Flushed incoming audio playback`. Local GStreamer backend barge-in unaffected.


andimarafioti (commented):
Looking good 👍 
My only issue here is that this needs a new SDK for it to make sense, right? So I would suggest we wait to merge until we can also update the pyproject.toml to the new sdk. wdyt?

FabienDanieau:
We can wait for the https://github.com/pollen-robotics/reachy_mini/pull/1186 to be merged an released, but the changes here should not break anything.

andimarafioti:
@copilot resolve the merge conflicts in this pull request

andimarafioti (approved):
Great!

PR #390, opened 2026-06-04
Add long-term memory tools

andimarafioti:
## Summary
- Add a small JSON-backed long-term memory store compatible with the mobile app memory shape.
- Add `remember` and `forget` built-in tools and enable them in bundled production profiles.
- Prepend saved memory facts to session instructions for OpenAI, Hugging Face, and Gemini backends.
- Document the new tools and add focused memory tests.

## Validation
- `uv run ruff check ...`
- `uv run mypy src/`
- `uv run pytest`


alozowski:
removed two unused pieces from the new memory module before merging:
- `MEMORY_PATH_ENV` was an env-var override for the memory file's location, but it never made it into use – nothing in the codebase set it or read it. It's also redundant: the actual path resolution already works correctly through `instance_path`, so this was just extra surface area with no purpose
- `fact_id` on `forget_memory_fact` was meant to let callers remove a fact by its internal ID, but its only caller, the forget tool, only ever passes a natural-language query, because the model has no way of knowing a fact's internal id. So this code path was never reachable

alozowski:
Also, added the clarification how to reset all memory into README

PR #391, opened 2026-06-04
Fix realtime tool result prompting

andimarafioti:
## What changed

- Remove the post-tool `response.instructions` override from the production tool-result path.
- Trigger non-idle tool follow-up responses with a bare `response.create()` so session instructions are not replaced for that model call.
- Keep the regression coverage focused on that behavior with one OpenAI realtime test.
- Merge `main` into the branch to resolve the PR conflicts.

## Why

The previous tool-result path sent a short `response.instructions` value after non-idle tool calls. In Realtime, that field overrides the response/session instructions for that response rather than appending a small hint. Calling `response.create()` without per-response instructions preserves the configured session prompt while still letting the model use the function-call output already present in the conversation.

## Validation

- `uv run pytest tests/test_openai_realtime.py -q` (25 passed)
- `uv run ruff check src/reachy_mini_conversation_app/base_realtime.py tests/test_openai_realtime.py`
- `uv run ruff format --check src/reachy_mini_conversation_app/base_realtime.py tests/test_openai_realtime.py`

Full-suite note: `uv run pytest` currently stops during collection because `tests/vision/test_local_vision.py` imports `torch`, which is not installed in this fresh worktree environment.


alozowski (approved):
Looks good!

PR #396, opened 2026-06-05
Reload tools when profiles change

andimarafioti:
## Summary

This fixes runtime personality/profile switches so the active tool registry follows the newly selected profile without requiring a Python process restart.

## Root Cause

`initialize_tools()` only populated the registry once per process. When the UI switched from a profile like `example` where `camera` is disabled to `default`/`tedai` where `camera` is enabled, the realtime session restarted but kept the old tool list. Profile-local tools that had already been imported could also leak into later profiles.

## Changes

- Add an active tool-registry signature based on profile, profile root, external tools root, autoload setting, and instance path.
- Keep a lazy cache of loaded tool classes keyed by source (`module`, `file`, or remote tool identity).
- Rebuild the active registry as a filtered view over only the sources requested by the current profile.
- Add regressions for `example -> default` tool reload behavior and forced profile-local reloads.

## Validation

- `pytest tests/test_external_loading.py tests/test_tool_space_runtime.py -q`
- `pytest tests -q --ignore=tests/test_gemini_live.py --ignore=tests/test_tool_spaces.py`
- `python -m ruff check src/reachy_mini_conversation_app/tools/core_tools.py tests/test_external_loading.py`
- `python -m compileall -q src/reachy_mini_conversation_app/tools/core_tools.py tests/test_external_loading.py`

Note: the full unignored suite is blocked in this local environment by missing `google.genai` and an installed `reachy_mini` package that does not expose `SetVolumeCmd`.

### DIFF
--- a/src/reachy_mini_conversation_app/tools/core_tools.py
+++ b/src/reachy_mini_conversation_app/tools/core_tools.py
@@ -79,8 +79,6 @@ ALL_TOOLS: Dict[str, Tool] = {}
 ALL_TOOL_SPECS: List[Dict[str, Any]] = []
 _TOOLS_INITIALIZED = False
 _TOOLS_SIGNATURE: tuple[str, str, str | None, bool, str | None] | None = None
-_TOOLS_SIGNATURE_CACHE_INPUTS: tuple[str, str | Path, str | Path | None, bool, str | Path | None] | None = None
-_TOOLS_SIGNATURE_CACHE: tuple[str, str, str | None, bool, str | None] | None = None
 _TOOLS_INSTANCE_PATH: str | Path | None = None
 _LOADED_TOOL_CLASS_CACHE: Dict[tuple[str, str], List[type[Tool]]] = {}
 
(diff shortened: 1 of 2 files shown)


src/reachy_mini_conversation_app/tools/core_tools.py
```diff
@@ -87,6 +78,10 @@ async def __call__(self, deps: ToolDependencies, **kwargs: Any) -> Dict[str, Any
 ALL_TOOLS: Dict[str, Tool] = {}
 ALL_TOOL_SPECS: List[Dict[str, Any]] = []
 _TOOLS_INITIALIZED = False
+_TOOLS_SIGNATURE: tuple[str, str, str | None, bool, str | None] | None = None
+_TOOLS_INSTANCE_PATH: str | Path | None = None
+ToolCacheKey = tuple[str, str]
```
alozowski: This ends up on the public surface of the module, it worth prefixing with `_`, or just inlining `tuple[str, str]` in the three annotation sites since the alias doesn't save much

src/reachy_mini_conversation_app/tools/core_tools.py
```diff
@@ -380,22 +467,35 @@ def _load_profile_tools(tool_names: list[str], remote_tool_names: set[str]) -> N
...
 
+    return loaded_tool_classes
 
-def initialize_tools(instance_path: str | Path | None = None) -> None:
-    """Populate the tool registry once for the current process."""
-    global ALL_TOOLS, ALL_TOOL_SPECS, _TOOLS_INITIALIZED
 
-    if _TOOLS_INITIALIZED:
-        logger.debug("Tools already initialized; skipping reinitialization.")
+def initialize_tools(instance_path: str | Path | None = None, *, force: bool = False) -> None:
```
alozowski, line 472: `force=True` re-executes profile-local `.py` files but not built-in tool modules, because those go through `importlib.import_module` which just returns the cached `sys.modules` entry. Worth a note in the docstring

src/reachy_mini_conversation_app/tools/core_tools.py
```diff
@@ -204,6 +269,17 @@ def _build_tool_registry(
     return {tool.name: tool for tool in tool_instances}
 
 
+def _tool_registry_signature(instance_path: str | Path | None) -> tuple[str, str, str | None, bool, str | None]:
```
alozowski, line 271: `Path.resolve()` is called on every tool dispatch to compute the signature, even on the fast path, before the early-return check. Should we cache it at startup?

alozowski (approved):
Agree with these changes! I left a few small comments, nothing blocking

PR #398, opened 2026-06-08
Reduce emotion tool schema size

andimarafioti:
## Summary

- replace the `play_emotion` tool schema full recorded-move catalog with a compact emotional-intent enum
- resolve compact intents, exact curated recorded move IDs, and exposed yes/no phrase variants locally before queueing the emotion
- preserve direct nuanced variants like `no_sad`, `no_excited`, `no_firm`, and `yes_understanding`
- remove bad-quality moves and substitute-only intents such as `yes_sad`, `yes_proud`, `no_confused`, `oops`, `lost`, `proud`, `curious`, `inquiring`, and `uncomfortable`
- avoid weak fallback moves when a precise recorded move is unavailable, so intents like `happy`, `grateful`, `no_sad`, and `yes_understanding` do not degrade into loosely related motions
- unknown explicit values fall back to a random curated recorded emotion
- intentionally do not maintain synonym or multilingual alias matching

## Prompt size impact

- Before: 9807 characters for `PlayEmotion().spec()`
- Now: 927 characters for `PlayEmotion().spec()`
- Reduction: 8880 characters, 90.5%

## Validation

- `uv run ruff check src/reachy_mini_conversation_app/tools/play_emotion.py tests/tools/test_play_emotion.py`
- `uv run pytest tests/tools/test_play_emotion.py`
- `uv run pytest tests/test_huggingface_realtime.py tests/test_openai_realtime.py tests/tools/test_play_emotion.py`


src/reachy_mini_conversation_app/tools/play_emotion.py
```diff
@@ -65,20 +276,18 @@ async def __call__(self, deps: ToolDependencies, **kwargs: Any) -> Dict[str, Any
...
             if not emotion_names:
                 return {"error": "No emotions currently available"}
 
+            emotion_name = resolve_emotion_name(requested_emotion, emotion_names)
             if not emotion_name:
-                emotion_name = random.choice(emotion_names)
-
-            if emotion_name not in emotion_names:
-                return {"error": f"Unknown emotion '{emotion_name}'. Available: {emotion_names}"}
+                emotion_name = random_curated_emotion(emotion_names)
```
RemiFabre, line 291: imo it's important to log when something different than what was requested is played, e.g.
logger.info("play_emotion: %r did not resolve; using random curated", requested_emotion)

RemiFabre (approved):
I suspect the loss in emotion context is compensated by the gain in delay. In my test the robot reacted quite well. Nice improvement!

src/reachy_mini_conversation_app/tools/play_emotion.py
andimarafioti: agreed, added it!

PR #401, opened 2026-06-11
Fix coalesce tool followup responses

### LINKED ISSUE #343: Batch tool result follow-up responses to avoid duplicate assistant turns
## Problem

When the local Hugging Face realtime backend receives a model response with multiple tool calls, the app currently appears to trigger a follow-up `response.create` once per completed tool result. That can make the assistant speak the same first turn twice.

## Observed behavior

Running `speech-to-speech --log_level debug` with the conversation app connected:

1. User says: `Hey Richie, how are you doing?`
2. The speech-to-speech server sends one assistant response containing only tool calls:
   - `play_emotion({"emotion":"attentive1"})`
   - `do_nothing({"reason":"listening to the client"})`
3. The app executes both tools and sends both `function_call_output` items.
4. Two separate follow-up `response.create` events are sent after the tool results.
5. The assistant says the same text twice:
   - `I'm doing well, thanks for asking. How about you?`
   - `I'm doing well, thanks for asking. How about you?`

The later request history also shows two duplicate assistant messages in context, confirming it was not a VAD duplicate of the user turn.

## Relevant log evidence

From the app logs:

```text
15:09:43 Tool call received — tool_name='play_emotion', call_id=call_93...
15:09:43 Tool call received — tool_name='do_nothing', call_id=call_d8...
15:09:43 Tool 'play_emotion' executed successfully.
15:09:43 Tool 'do_nothing' executed successfully.
15:09:44 role=assistant content=I'm doing well, thanks for asking. How about you?
15:09:47 role=assistant content=I'm doing well, thanks for asking. How about you?
```

From the speech-to-speech logs:

```text
15:09:43 response.create received, LLM generation triggered
15:09:46 response.create received, LLM generation triggered
```

## Expected behavior

For one assistant response containing multiple tool calls, the app should:

1. Execute all tool calls from that assistant response.
2. Send all corresponding `function_call_output` items.
3. Trigger exactly one follow-up `response.create` for the batch.

## Likely fix area

`src/reachy_mini_conversation_app/base_realtime.py`, around `_handle_tool_result`.

Currently each non-idle tool result can call `_safe_response_create(...)` independently. That should be batched/debounced per assistant response or per pending tool-call group, so multiple tool outputs from the same model response cause only one follow-up response.

## Notes

`speech-to-speech` is behaving like a realtime server here: every received `response.create` means generate a response. The cleaner protocol fix is in the client/app side rather than having the server coalesce duplicate generation requests.

alozowski:
## Summary
Fixes #343. Two related causes of duplicate/unwanted spoken assistant turns after tool calls:
- First problem: when a robot turn contains multiple parallel tool calls (for example, two calls to
  `remember`), each completed tool independently queued its own `response.create`, causing the assistant
  to repeat the same reply 2-3 times. `_response_sender_loop` now coalesces queued empty-kwargs follow-up
  requests into one
- Second problem: tools that are pure physical side effects (`dance`, `stop_dance`, `play_emotion`,
  `stop_emotion`, `move_head`) no longer trigger a spoken follow-up at all. For example, `stop_dance` no
  longer causes the model to add an unprompted "Got it!" after stopping. New `needs_response: ClassVar[bool]`
  on `Tool` (default `True`), set to `False` for these tools; `_handle_tool_result` checks it before
  calling `response.create`, while errors always still surface

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [ ] Head wobble
</details>

### DIFF
--- a/src/reachy_mini_conversation_app/base_realtime.py
+++ b/src/reachy_mini_conversation_app/base_realtime.py
@@ -33,8 +33,8 @@ from reachy_mini_conversation_app.config import (
 )
 from reachy_mini_conversation_app.idle_policy import start_idle_tool_call
 from reachy_mini_conversation_app.tools.core_tools import ToolDependencies
-from reachy_mini_conversation_app.tools.tool_constants import SILENT_TOOLS
 from reachy_mini_conversation_app.conversation_handler import ConversationHandler
+from reachy_mini_conversation_app.tools.tool_constants import SILENT_TOOLS
 from reachy_mini_conversation_app.tools.background_tool_manager import (
     ToolCallRoutine,
     ToolNotification,
--- a/src/reachy_mini_conversation_app/base_realtime.py
+++ b/src/reachy_mini_conversation_app/base_realtime.py
@@ -26,6 +26,7 @@ from openai.types.realtime import (
 from websockets.exceptions import ConnectionClosedError
 from openai.resources.realtime.realtime import AsyncRealtimeConnection
 
+from reachy_mini_conversation_app.tools import core_tools
 from reachy_mini_conversation_app.config import (
     config,
     get_default_voice_for_backend,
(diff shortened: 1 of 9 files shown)


andimarafioti (approved):
Love this! Thank you 🙏

PR #403, opened 2026-06-11
Update realtime voices in place and fix tool result ordering

andimarafioti:
## Summary

- Update realtime voice changes in place via `session.update` instead of rebuilding the active backend.
- Keep personality changes owned by the backend restart path, while preserving safer validation, playback flushing, and failed-apply handling.
- Delay realtime tool result submission until the tool-call response has completed, fixing camera/tool outputs that could be sent before the backend had recorded the function call.

## Details

- `LocalStream.change_voice()` now delegates to the active handler, and OpenAI-compatible realtime handlers send an `audio.output.voice` session update when connected.
- Personality applies still validate the requested profile and then request a backend restart; the removed same-endpoint reconnect path is no longer part of this PR.
- Failed personality resolution rolls back the previously selected profile, and the headless UI reports failed applies as `ok: false` so they are not persisted as startup settings.
- Tool results now wait for `response.done` before sending `function_call_output`; if the wait times out, the app surfaces the tool result locally but does not submit an invalid model result.

## Validation

- `.venv/bin/pytest tests/test_console.py tests/test_huggingface_realtime.py tests/test_openai_realtime.py tests/test_gemini_live.py -q` (`79 passed`)
- `.venv/bin/ruff check src/reachy_mini_conversation_app/base_realtime.py src/reachy_mini_conversation_app/console.py src/reachy_mini_conversation_app/gemini_live.py src/reachy_mini_conversation_app/headless_personality_ui.py tests/test_console.py tests/test_gemini_live.py tests/test_huggingface_realtime.py tests/test_openai_realtime.py tests/vision/test_yolo_process.py`
- `.venv/bin/mypy`
- `.venv/bin/pytest -q` (`258 passed`)


A-Mahla (changes requested):
Thanks @andimarafioti, good for me. Much better to update instead to restart a new connection 👍 . Just codex catched a Gemini related issue :

Codex finding:
1. High: Gemini failed personality applies no longer roll back the selected profile.
     Th PR moved LocalStream.apply_personality() to delegate directly to the active handler at
     console.py:409

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/console.py#L409).
     OpenAI/HF rollback is implemented in BaseRealtimeHandler, but
     GeminiLiveHandler.apply_personality() sets the profile at gemini_live.py:245

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/gemini_live.py#L245)
     and returns failure at gemini_live.py:253

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/gemini_live.py#L253)
     without restoring the previous profile. This contradicts the PR body claim that failed
     personality resolution rolls back the previous selected profile, and it is a real behavioral
     regression for Gemini.

  2. Medium: failed personality applies are returned as JSON ok: false, but the bundled frontend
     still treats them as success.
     The route now computes apply_ok and returns it at headless_personality_ui.py:273

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/headless_personality_ui.py#L273),
     but static/main.js only checks HTTP status at main.js:241

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/static/main.js#L241)
     and then displays the returned status as success at main.js:890

     (https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/ff9a1af/src/reachy_mini_conversation_app/static/main.js#L890).
     External callers may also notice this API contract change: failure is now represented in JSON
     while still using HTTP 200.

A-Mahla (approved):

andimarafioti:
Superseded by #406 and #407.

PR #404, opened 2026-06-12
feat: audio is boosted

Toon-nooT:
## Summary

Add audio output gain control to the conversation app. This applies a configurable linear gain multiplier (0–24 dB) to outgoing audio frames before they reach the robot speaker, making the robot's voice louder without requiring hardware volume changes. Includes:

- `audio_output_gain.py` module for per-frame gain processing
- `audio/startup_config.py` for tuned XVF3800 audio parameters at startup (AGC, noise suppression, echo cancellation)
- REST API endpoints (`GET/POST /api/audio/output_gain`) for runtime gain adjustment
- Handler hot-swap with `handler_factory` and backend reconnection state machine

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [x] `.env.example` updated (if new config vars were added)
- [x] Installation from the desktop app tested (if applicable)

<details>

- [x] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [ ] Head wobble
</details>

## Notes

Tested via the desktop app in headless mode. The gain is applied per-frame in the play loop so changes take effect on the next speech output without requiring a restart. The startup config tunes the XVF3800 DSP parameters (AGC max gain, noise suppression, echo cancellation) for better conversation audio quality.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -45,3 +45,7 @@ HF_TOKEN=
 # even if they are not listed in the selected profile's tools.txt.
 # This is convenient for downloaded tools used with built-in/default profiles.
 # AUTOLOAD_EXTERNAL_TOOLS=1
+
+# App-level audio output gain (0–24 dB). Boosts speaker volume beyond system default.
+# Useful in noisy environments. 0 = no boost. Applied per-frame, no restart needed.
+# FAMILIAR_AUDIO_GAIN_DB=0.0
--- a/src/reachy_mini_conversation_app/audio_output_gain.py
+++ b/src/reachy_mini_conversation_app/audio_output_gain.py
@@ -1,0 +1,66 @@
+"""App-level audio output gain control.
+
+Applies a linear gain multiplier to outgoing audio frames before they reach
+the robot speaker.  The value is read per-frame in play_loop() so changes
+take effect on the next speech output without a restart.
+"""
+
+import os
+
+_DEFAULT_GAIN_DB = 0.0
+_MIN_GAIN_DB = 0.0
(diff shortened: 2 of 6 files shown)


andimarafioti (changes requested):
This a fair feature and I think it would be useful, but the audio engineer in me can't accept a compressor with a fix gain rate that just clamps on (-1, 1). I think we could do better by applying the gain only if the signal is below a threshold and making sure the upper limit is less drastic, ie, you can still understand the signal when it goes over.

Toon-nooT:
@andimarafioti From your expertise, would it make more sense to replace np.clip with np.tanh? From what i gather, a tanh curve naturally compresses peaks toward ±1.0 without ever reaching it, preserving the waveform shape more. I could let it only engage above a threshold, leaving everything below it perfectly linear (untouched).
Or do you have other suggestions?

andimarafioti:
A tanh would be better, but generally, I would read a bit into limiters and compressors for a "good sounding" approach. Besides that, there is another level which we don't control: the robot's volume itself. We might want to play with that as well.

FabienDanieau:
please look at https://github.com/pollen-robotics/reachy_mini/issues/569#issuecomment-4796927329 for a first solution

Toon-nooT:
See my comment here: https://github.com/pollen-robotics/reachy_mini/pull/1187#issuecomment-4833390218
The firmware update boosts audio, this PR is unnecessary

PR #405, opened 2026-06-15
Add startup greeting prompt

andimarafioti:
## Summary

Adds a profile-driven startup greeting so Reachy opens the conversation once the realtime backend is connected and ready.

## Changes

- Adds optional `greeting.txt` profile support with a default prompt.
- Queues a first internal text turn for OpenAI-compatible realtime sessions and Gemini Live.
- Adds greeting load/save support to Gradio and headless personality UIs.
- Documents the new profile file and covers the behavior in tests.

## Validation

- `.venv/bin/pytest tests/test_openai_realtime.py tests/test_huggingface_realtime.py tests/test_gemini_live.py tests/test_profile_paths.py`
- `.venv/bin/ruff check src/reachy_mini_conversation_app/prompts.py src/reachy_mini_conversation_app/base_realtime.py src/reachy_mini_conversation_app/gemini_live.py src/reachy_mini_conversation_app/headless_personality.py src/reachy_mini_conversation_app/headless_personality_ui.py src/reachy_mini_conversation_app/gradio_personality.py tests/test_profile_paths.py tests/test_openai_realtime.py tests/test_gemini_live.py`

### DIFF
--- a/profiles/default/greeting.txt
+++ b/profiles/default/greeting.txt
@@ -1,1 +1,0 @@
-Start the conversation now with a brief, spontaneous greeting in character. Keep it to one sentence, invite the user in naturally, and vary the wording each time.
--- a/profiles/default/greeting.txt
+++ b/profiles/default/greeting.txt
@@ -1,0 +1,1 @@
+Start the conversation now with a brief, spontaneous greeting in character. Keep it to one sentence, invite the user in naturally, and vary the wording each time.
--- a/src/reachy_mini_conversation_app/gemini_live.py
+++ b/src/reachy_mini_conversation_app/gemini_live.py
@@ -31,7 +31,11 @@ from reachy_mini_conversation_app.config import (
     DEFAULT_VOICE_BY_BACKEND,
     config,
 )
-from reachy_mini_conversation_app.prompts import get_session_voice, get_session_instructions
+from reachy_mini_conversation_app.prompts import (
+    get_session_voice,
+    get_session_instructions,
+    get_session_greeting_prompt,
+)
 from reachy_mini_conversation_app.idle_policy import start_idle_tool_call
 from reachy_mini_conversation_app.tools.core_tools import (
     ToolDependencies,
@@ -182,6 +186,7 @@ class GeminiLiveHandler(ConversationHandler):
         self._pending_user_transcript_chunks: list[str] = []
         self._pending_assistant_transcript_chunks: list[str] = []
         self._listening_state = False
+        self._startup_greeting_sent = False
 
     def copy(self) -> "GeminiLiveHandler":
         """Return a fresh handler, preserving deps and voice override."""
@@ -437,6 +442,35 @@ class GeminiLiveHandler(ConversationHandler):
 
             logger.info("Started background tool: %s (id=%s, call_id=%s)", tool_name, bg_tool.tool_id, call_id)
 
+    async def _send_startup_greeting_prompt(self) -> None:
+        """Prompt Gemini to open the conversation once the live session is ready."""
+        if self._startup_greeting_sent or not self.session:
+            return
+
+        greeting_prompt = get_session_greeting_prompt(self.instance_path).strip()
+        if not greeting_prompt:
+            self._startup_greeting_sent = True
+            return
+
+        send_client_content = getattr(self.session, "send_client_content", None)
+        if not callable(send_client_content):
+            logger.warning("Gemini session does not support send_client_content; startup greeting skipped")
+            return
+
+        try:
+            await send_client_content(
+                turns=types.Content(
+                    role="user",
+                    parts=[types.Part(text=greeting_prompt)],
+                ),
+                turn_complete=True,
+            )
+            self._startup_greeting_sent = True
+            self.last_activity_time = time.monotonic()
+            logger.info("Queued Gemini startup greeting prompt")
+        except Exception as e:
+            logger.warning("Failed to queue Gemini startup greeting prompt: %s", e)
+
     async def _handle_tool_result(self, bg_tool: ToolNotification) -> None:
         """Process the result of a completed tool and send it back to Gemini."""
         if bg_tool.error is not None:
@@ -548,6 +582,8 @@ class GeminiLiveHandler(ConversationHandler):
                 # Start video sender if camera is available
                 if self.deps.camera_worker is not None:
                     video_task = asyncio.create_task(self._video_sender_loop(), name="gemini-video-sender")
+
+                await self._send_startup_greeting_prompt()
 
                 # session.receive() yields responses for the current turn then completes.
                 # We loop so the session stays alive across multiple conversation turns.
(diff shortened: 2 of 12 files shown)


profiles/default/greeting.txt
```diff
@@ -0,0 +1 @@
+Start the conversation now with a brief, spontaneous greeting in character. Keep it to one sentence, invite the user in naturally, and vary the wording each time.
```
alozowski: This is byte-identical to the `DEFAULT_GREETING_PROMPT` constant in `prompts.py`, which is already the fallback when the file is missing. Could we drop the file and let the constant be the single source of truth?

src/reachy_mini_conversation_app/prompts.py
```diff
@@ -114,3 +120,18 @@ def get_session_voice(default: str | None = None) -> str:
     except Exception:
         pass
     return fallback
+
+
+def get_session_greeting_prompt(instance_path: str | Path | None = None) -> str:
```
alozowski: this function never uses the arg (_ = instance_path), so it looks like it's mirroring `get_session_instructions`, but that one actually uses it. I suggest dropping the param and the two call sites that pass it

src/reachy_mini_conversation_app/static/js/components/profile-modal.js
```diff
@@ -180,6 +181,21 @@ function buildDialog({ isEdit, initial, toolChoices, isToolChecked }) {
...
+        h("span", { class: "modal__label" }, "Startup greeting prompt"),
+        h(
+          "textarea",
+          {
+            name: "greeting",
+            rows: "3",
+            placeholder: "Start the conversation with a short greeting in character.",
+            class: "modal__textarea",
+          },
+          initial.greeting || "Start the conversation with one short greeting in character. Vary the wording each time."
```
alozowski: The greeting box prefils a hardcoded default value, while the instructions box right above uses `initial.instructions || ""` + a placeholder. What if we match that here (`initial.greeting || ""`)? The placeholder on line 193 already gives the hint, and it avoids saving this wording as if the user wrote it

src/reachy_mini_conversation_app/gemini_live.py
```diff
@@ -425,6 +430,35 @@ async def _handle_tool_call(self, response: Any) -> None:
...
+        """Prompt Gemini to open the conversation once the live session is ready."""
+        if self._startup_greeting_sent or not self.session:
+            return
+
+        greeting_prompt = get_session_greeting_prompt(self.instance_path).strip()
+        if not greeting_prompt:
+            self._startup_greeting_sent = True
+            return
+
+        send_client_content = getattr(self.session, "send_client_content", None)
```
alozowski, line 443: When `send_client_content` isn't available we return without setting `_startup_greeting_sent = True`, so every reconnect re-logs the same warning. Since it's a permanent capability, setting the flag first would make it a one-time thing

src/reachy_mini_conversation_app/prompts.py
```diff
@@ -114,3 +120,18 @@ def get_session_voice(default: str | None = None) -> str:
...
+    """Resolve the startup greeting prompt for the selected profile."""
+    _ = instance_path
+    profile = config.REACHY_MINI_CUSTOM_PROFILE or "default"
```
alozowski: For the no-profile case, greeting reads from `profiles/default/`, but instructions read from the `prompts/` library. These are slightly different roots for "the default" and it works today, and becomes moot if we drop the duplicate file above (just noting it)

src/reachy_mini_conversation_app/personality_routes.py
```diff
@@ -130,6 +133,7 @@ async def _save(request: Request) -> dict:  # type: ignore
             raw = {}
         name = str(raw.get("name", ""))
         instructions = str(raw.get("instructions", ""))
+        greeting = str(raw["greeting"]) if raw.get("greeting") is not None else None
```
alozowski, line 136: Greeting parses with a `None` sentinel so `_write_profile` can skip creating `greeting.txt`, while `instructions/tools/voice` use `.get(..., "")` and always write. It's a minor asymmetry, just flagging it

alozowski (commented):
I like this feature! I left some code review comments and I'm going to test it on a real robot in various scenarios 

One general question, should switching personality re-greet? I noticed that `_startup_greeting_sent` lasts the whole handler lifetime, and `_restart_session` reuses the same instance in both backends, so the flag stays `True` across a restart. That means a reconnect won't double-greet, which is correct, but switching personality at runtime also stays silent. Both backends do the same thing. If a new personality should say hello, resetting the flag in `apply_personality` would do it, but it's a design question

profiles/default/greeting.txt
andimarafioti: Done in `4523762`: removed `profiles/default/greeting.txt` so `DEFAULT_GREETING_PROMPT` is the single source of truth for the built-in fallback.

src/reachy_mini_conversation_app/prompts.py
andimarafioti: Done in `4523762`: removed the unused `instance_path` parameter from `get_session_greeting_prompt()` and updated the OpenAI/Gemini call sites and tests.

src/reachy_mini_conversation_app/static/js/components/profile-modal.js
andimarafioti: Done in `4523762`: the modal now uses `initial.greeting || ""`. I also stopped the create flow from substituting the built-in fallback before save, so blank means “use the built-in default”.

src/reachy_mini_conversation_app/gemini_live.py
andimarafioti: Good catch, fixed in `4523762`: the unsupported `send_client_content` path now sets `_startup_greeting_sent = True` before returning, and there is a regression test that confirms the warning is emitted only once.

src/reachy_mini_conversation_app/prompts.py
andimarafioti: Yep, this became moot after dropping the duplicate file in `4523762`. The no-profile greeting path now returns `DEFAULT_GREETING_PROMPT` directly instead of resolving `profiles/default/greeting.txt`.

src/reachy_mini_conversation_app/personality_routes.py
andimarafioti: I kept the greeting sentinel intentionally and tightened the behavior around it in `4523762`: omitted/empty greeting now avoids creating `greeting.txt` and removes an existing one on save, so runtime falls back to the built-in default. Added tests for the empty custom greeting path.

andimarafioti:
Thanks for testing this on the robot. On the personality-switching question: I left the behavior unchanged in this PR. My read is that reconnects should stay silent, and personality switches/edits should not auto-greet unless we make that an explicit UX decision; otherwise changing a setting could unexpectedly start a new spoken turn. If we do want the new personality to say hello, resetting the flag when applying a new personality should be a small follow-up.

alozowski (approved):
Tested, it works good!

PR #406, opened 2026-06-16
Fix realtime tool result ordering

andimarafioti:
## Summary

- Wait for `response.done` before submitting realtime `function_call_output`.
- Drop model-facing tool results when response completion times out, while still surfacing the tool result locally.
- Stabilize the active-response retry test so the rejection path is deterministic.

## Validation

- `.venv/bin/pytest tests/test_openai_realtime.py -q` (`28 passed`)
- `.venv/bin/ruff check src/reachy_mini_conversation_app/base_realtime.py tests/test_openai_realtime.py`
- `.venv/bin/mypy`


A-Mahla (approved):
Nice, @andimarafioti! Good catch. This should avoid a lot of retries caused by the “conversation already has a response” error.

The only edge case is that the robot could have run the tool but never received its result. But the 30-second timeout between the function_tool_call event and response.done should prevent that kind of situation.

PR #407, opened 2026-06-16
Update realtime voices in place

andimarafioti:
## Summary

- Change OpenAI-compatible realtime voices with `session.update` instead of restarting the active backend.
- Delegate settings UI voice changes to the active handler so the selected backend stays connected.
- Keep the selected voice override synchronized after a successful handler update.

## Validation

- `.venv/bin/pytest tests/test_console.py tests/test_huggingface_realtime.py -q` (`39 passed`)
- `.venv/bin/ruff check src/reachy_mini_conversation_app/base_realtime.py src/reachy_mini_conversation_app/console.py tests/test_console.py tests/test_huggingface_realtime.py`
- `.venv/bin/mypy`

### DIFF
--- a/src/reachy_mini_conversation_app/base_realtime.py
+++ b/src/reachy_mini_conversation_app/base_realtime.py
@@ -274,16 +274,27 @@ class BaseRealtimeHandler(ConversationHandler, ABC):
         )
 
     async def change_voice(self, voice: str) -> str:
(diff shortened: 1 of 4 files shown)


A-Mahla (approved):
LGTM! thanks @andimarafioti

PR #409, opened 2026-06-17
Hide always-injected and support tool modules from personality UI

scienceetonnante:
Some tools are exposed in personality UI but shouldn't:

- System tools are always present (so shouldn't be exposed with a checkbox)
- same for support tool modules "background_tool_manager", "tool_constants"

This PR simply removes them from the list exposed tools in Gradio and headless personality UI


andimarafioti (approved):
Thank you! 🙏

PR #410, opened 2026-06-17
Add app inactivity timeout

andimarafioti:
## Summary

- Add an app inactivity timeout that closes the conversation app after a configured period without conversation activity.
- Support `--app-timeout-minutes` and `REACHY_MINI_APP_TIMEOUT_MINUTES`; unset values default to `1440` minutes, while zero or negative values disable the timeout.
- Reuse the handlers' existing `last_activity_time` for app timeout checks, and track idle behavior cadence separately with `last_idle_behavior_time` so idle actions do not keep the app alive.

## Validation

- `.venv/bin/pytest tests/test_main.py tests/test_utils.py -q` (`11 passed`)
- `.venv/bin/ruff check src/reachy_mini_conversation_app/main.py src/reachy_mini_conversation_app/utils.py tests/test_main.py tests/test_utils.py`
- `.venv/bin/ruff format --check src/reachy_mini_conversation_app/main.py src/reachy_mini_conversation_app/utils.py tests/test_main.py tests/test_utils.py`


alozowski:
I did some refactoring, here is what was changed:
- dropped the `--app-timeout-minutes flag` because it duplicated the env var, wasn't documented, and couldn't be set when the app runs via `ReachyMiniConversationApp.run()`, so I kept `REACHY_MINI_APP_TIMEOUT_MINUTES` as the one knob
- moved env parsing into config.py, that's where every other `REACHY_MINI_*` var lives, so config stays in one place as it was designed
- replaced `Any`/`getattr` with real types, reading activity via a small `LocalStream.seconds_since_activity()` accessor instead of poking handler internals
- declared `last_activity_time` on the handler base so it's a shared contract

PR #415, opened 2026-06-19
Unify idle behavior

### LINKED ISSUE #412: Unify idle/activity handling across the realtime handlers
The idle-behavior loop and the activity/idle timestamps are implemented twice: once in `BaseRealtimeHandler` (OpenAI/Hugging Face) and again in `GeminiLiveHandler`, and the copies differ (thresholds of 180s vs 15s, and the realtime base guards on `_response_done_event` while Gemini doesn't).

Any change to idle or activity logic has to be made in both places and kept in sync, which is easy to get wrong and makes the overall idle behavior hard to reason about. This came up while adding the app inactivity timeout, which had to add `last_idle_behavior_time` to both handlers.

alozowski:
## Summary
Fixes https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/412
The idle-behavior loop and the activity/idle timestamps were implemented twice: once in `BaseRealtimeHandler` (OpenAI / Hugging Face) and again in `GeminiLiveHandler`, and the copies had drifted, so any change to idle or activity logic had to be made in both places and kept in sync. This came up while adding the app inactivity timeout, which had to add the same state to both handlers.
- Moved the idle loop, the activity/idle timestamps, `_mark_activity`, `emit` and the idle dispatch into the shared `ConversationHandler` – one source of truth that can't drift, so the inactivity timeout (and any future idle change) is made once. Each backend only provides two hooks for its real differences: `_is_connected` (connection vs session) and `_idle_behavior_ready` (the response-in-progress guard the realtime base needs and Gemini doesn't). The idle threshold is standardised to 180s for all backends (Gemini was 15s) so behavior is consistent and tuned in one place, and the activity clock to a single `time.monotonic()` source (Gemini compared `monotonic()` against an asyncio-loop timestamp, which matched only by coincidence)
- Typed tool specs with a `ToolSpec` TypedDict instead of `dict[str, Any]`: the shape is now checked at type-check time, which made the runtime "is this a valid spec" checks dead, so they're removed
- Renamed the obscure `bg_tool`, which named two different objects, to `background_tool` (the running handle) and `completed_tool` (the finished-tool notification) so the name tells you which one it is

## Category
- [ ] Fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Headless mode (default)
- [ ] Gradio UI (`--gradio`)
</details>

<details>

- [ ] Head wobble
</details>

### DIFF
--- a/src/reachy_mini_conversation_app/base_realtime.py
+++ b/src/reachy_mini_conversation_app/base_realtime.py
@@ -7,11 +7,10 @@ import asyncio
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Final, Tuple, ClassVar, Optional
-from datetime import datetime
 
 import numpy as np
 from openai import AsyncOpenAI
-from fastrtc import AdditionalOutputs, wait_for_item, audio_to_int16
+from fastrtc import AdditionalOutputs, audio_to_int16
 from pydantic import Field, BaseModel
 from numpy.typing import NDArray
 from scipy.signal import resample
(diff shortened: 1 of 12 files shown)


andimarafioti (approved):
Good job!

PR #416, opened 2026-06-22
Rebuild the tool registry even when not changing personality

scienceetonnante:
When checking/unchecking tools without changing the personality, the tool registry was not rebuilt so the change of tools would not work.

This fix is to rebuild the tool registry.

Tested on both gradio and console interface.


alozowski (approved):
Looks good!

PR #356, opened 2026-05-13
Replace Gradio with a standalone conversation UI

tfrere:
## Summary

Adds an opt-in modern frontend for the headless settings server, served from a new `static_v2/` bundle. The legacy UI in `static/` is **untouched** and remains the default; the new UI is gated behind `REACHY_MINI_UI=modern` (and any user can revert with `REACHY_MINI_UI=legacy`).

The new UI brings two long-asked-for visuals into the conversation app:

- **Personality picker as a card grid**, using the avatar set originally introduced in `feat/new-ui` (13 SVGs). Selecting a card applies the personality and switches to the live conversation view.
- **Live conversation orb** that reacts to backend activity (`listening` / `thinking` / `speaking` / `idle`). The visual language mirrors `reachy_mini_minimal_conversation`'s orb so the three surfaces (mobile, minimal, conversation) read as the same widget across the product.

A small "Custom" card lets users create a quick personality (name + instructions) without touching files; the new profile gets the standard 8-tool set so the robot keeps its full expressivity.

### Backend (additive, zero regression by default)

| File | Change |
|---|---|
| `base_realtime.py` | New opt-in `set_activity_observer(callback)` API. The handler calls back on every `_mark_activity` transition so external surfaces can broadcast state without coupling. Defensive `try/except`, no impact when no observer is attached. |
| `conversation_events.py` (new) | `ConversationEventBus`: thread-safe pub/sub. Handler runs on its own asyncio loop, FastAPI serves SSE on another loop; the bus dispatches via `call_soon_threadsafe` and bounded queues so the realtime loop never blocks on slow clients. |
| `console.py` | `_select_ui_dir_name()` toggle + `_detach_framework_root_routes()` (the apps framework eagerly registers `/` and `/static` against the per-instance `static/` folder, this strips them so our routes can override while keeping URLs identical) + new `GET /conversation_events` SSE route with keep-alive and clean disconnect handling. Bus instantiated in `LocalStream.__init__` and attached to the handler. |
| `pyproject.toml` | `static_v2/**/*` added to `package-data` so the new bundle ships in the wheel. |

### Frontend (`src/reachy_mini_conversation_app/static_v2/`, all new)

11 ES modules + sectioned CSS + 13 SVG avatars, no toolchain. Each module has a single responsibility:

```
static_v2/
├── index.html                  # Shell HTML (header + outlet)
├── style.css                   # Sectioned single-file stylesheet (8 sections)
├── avatars/                    # 13 SVG avatars (LFS)
└── js/
[...]

PR #418, opened 2026-06-22
Add tool selection to the custom personality creation modal

scienceetonnante:
Restore per-personality tool selection that was lost when Gradio was replaced by the standalone UI (#356). 

The create-personality modal now shows a checklist of the available tools.

Changes are UI only.


andimarafioti (approved):
Great work!

PR #420, opened 2026-06-22
[codex] Avoid restarting unchanged personality

andimarafioti:
## Summary

- Skip the personality apply request in the UI when the selected card is already active.
- Make `/personalities/apply` idempotent for the active profile so direct API calls do not restart the realtime handler.
- Add console route regressions for same-profile apply and same-profile startup persistence.

## Root Cause

Selecting the active personality still sent a personality apply request. The backend always forwarded apply requests to the realtime handler, which rebuilt or restarted the active conversation session even though the profile had not changed.

## Impact

Users can browse personality selection and return to the current chat without ending or restarting the conversation, unless they actually choose a different personality.

## Validation

- `.venv/bin/pytest tests/test_console.py`

### DIFF
--- a/src/reachy_mini_conversation_app/personality_routes.py
+++ b/src/reachy_mini_conversation_app/personality_routes.py
@@ -178,7 +178,6 @@ def mount_personality_routes(
                 "ok": True,
                 "status": "Personality unchanged.",
                 "startup": persisted_choice,
-                "changed": False,
             }
 
         loop = get_loop()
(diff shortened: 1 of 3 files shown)


alozowski:
- removed a `changed` field from the apply response that no code ever read, the frontend already handles the no-op case directly, so the field was redundant
- renamed three confusing variables in the apply route (`sel_name`, `persist_flag`, `sel`) to clearer names

alozowski (approved):

PR #431, opened 2026-06-24
Remove fastrtc dependency

andimarafioti:
## Summary
Remove the `fastrtc` dependency by replacing the small subset used by the app with a local streaming helper. The helper keeps the existing queue timeout, additional text outputs, stream metadata, and audio dtype conversion behavior used by the realtime backends and local audio loop.

This also removes the orphaned `fastrtc` dependency stack from `uv.lock` and updates the README overview so it no longer names `fastrtc` as part of the runtime.

## Category
- [ ] Fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [x] `.env.example` updated (no new config vars were added)

<details>

- [x] Console mode (default)
- [ ] Web UI (`--ui`)
</details>

<details>

</details>

## Notes
Local clean-worktree gate passed with `UV_DEFAULT_INDEX=https://pypi.org/simple` to avoid the developer machine's uv registry override:

`ruff check . --fix && ruff format . && mypy --pretty --show-error-codes && pytest tests/ -v`

Result: 278 passed.

Architecture diagram update not needed; this removes a dependency shim without changing the app architecture.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,9 +11,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     #Media
-    "aiortc>=1.13.0",
     "av",
-    "fastrtc>=0.0.34",
     "huggingface-hub>=1.17.0",
     "httpx>=0.28.1",
 
--- a/src/reachy_mini_conversation_app/streaming.py
+++ b/src/reachy_mini_conversation_app/streaming.py
@@ -1,0 +1,102 @@
+import asyncio
+import warnings
+from typing import Literal, TypeVar, TypeAlias
+from collections.abc import Mapping, Callable
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+StreamMessage: TypeAlias = Mapping[str, object]
+AudioArray: TypeAlias = NDArray[np.int16] | NDArray[np.float32]
+AudioInput: TypeAlias = AudioArray | tuple[int, AudioArray]
+
+QueueItem = TypeVar("QueueItem")
+
+
+class AdditionalOutputs:
+    """Text or metadata emitted alongside audio frames."""
+
+    def __init__(self, *args: StreamMessage) -> None:
+        """Initialize with one or more emitted messages."""
+        self.args = args
+
+
+class AsyncStreamHandler:
+    """Minimal async stream handler state used by the local audio loop."""
+
+    def __init__(
+        self,
+        expected_layout: Literal["mono", "stereo"] = "mono",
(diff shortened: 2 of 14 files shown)


pyproject.toml
```diff
@@ -13,7 +13,6 @@ dependencies = [
     #Media
     "aiortc>=1.13.0",
```
FabienDanieau: Do we need this explicitly? I was included in fastrtc and I think It was manually bumped for security issues

pyproject.toml
```diff
@@ -13,7 +13,6 @@ dependencies = [
...
     "av",
```
FabienDanieau, line 14: We only use av for video conversion. Why not making reachy mini directly giving you the right format?

src/reachy_mini_conversation_app/streaming.py
```diff
@@ -0,0 +1,102 @@
...
+        warnings.warn(
+            "Passing a (sample_rate, audio) tuple is deprecated; pass only the audio array.",
+            UserWarning,
+            stacklevel=2,
+        )
+        return audio[1]
+    return audio
+
+
+def audio_to_int16(audio: AudioInput) -> NDArray[np.int16]:
```
FabienDanieau, line 85: Same than for video (last comment). Let's use gstreamer to directly give the right format on reachy mini side (so we may need to specify what we want if it's different from the default one)

pyproject.toml
andimarafioti: If it was only from fastrtc, I can remove it. I didn't know that.

pyproject.toml
andimarafioti: This seems a bit out-of-scope for this PR, isn't it?

src/reachy_mini_conversation_app/streaming.py
andimarafioti: I'll look into this, I was just exporitng the functionality from fastrtc

FabienDanieau: Ok so we could also use av for that (see https://huggingface.co/spaces/pollen-robotics/reachy_mini_radio/blob/main/reachy_mini_radio/main.py#L90). I'd like to avoid sound processing in python.
Then we'll see later to remove av

andimarafioti: Checked a few ways and the implementations were way more complex than this, so I left it

PR #439, opened 2026-06-29
Add a "delete" button to custom profiles

scienceetonnante:
The new UI allows to create and edit custom profiles personality, but not delete them.
This adds a small trash icon, a confirmation window.
It applies only to custom profiles (not built-in ones)
There is a guard against deleting the current active profile (icon does not show up).

image

### DIFF
--- a/src/reachy_mini_conversation_app/personality.py
+++ b/src/reachy_mini_conversation_app/personality.py
@@ -115,6 +115,24 @@ def available_tools_for(selected: str) -> List[str]:
     return sorted(set(shared + local))
 
 
+def delete_personality(name: str) -> bool:
+    """Delete a user-created personality directory; return whether it existed.
+
+    Refuses anything outside the user personalities root, so built-in profiles
+    can never be removed through the UI.
+    """
+    import shutil
+
+    target = resolve_profile_dir(name).resolve()
+    user_root = config.user_personalities_root().resolve()
+    if user_root not in target.parents:
+        return False
+    if target.is_dir():
+        shutil.rmtree(target)
+        return True
+    return False
+
+
 def _write_profile(
     sanitized_name: str,
     instructions: str,
--- a/src/reachy_mini_conversation_app/personality_routes.py
+++ b/src/reachy_mini_conversation_app/personality_routes.py
@@ -170,8 +170,20 @@ def mount_personality_routes(
     @app.delete("/personalities")
     def _delete(name: str) -> dict:  # type: ignore
         """Delete a user-created personality (name is the full selection string)."""
+        if name in (_current_choice(), _startup_choice()):
+            # Deleting the active/startup profile would break get_session_instructions() at next startup.
+            return JSONResponse(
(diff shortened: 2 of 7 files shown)


src/reachy_mini_conversation_app/personality_routes.py
```diff
@@ -166,6 +167,12 @@ async def _save(request: Request) -> dict:  # type: ignore
         except Exception as e:
             return JSONResponse({"ok": False, "error": str(e)}, status_code=500)  # type: ignore
 
+    @app.delete("/personalities")
```
alozowski, line 170: Could we guard this route against deleting the current or persisted startup profile? The project stores startup personality in `startup_settings.py`, then loads it back into runtime config, so if that profile directory is deleted, `prompts.get_session_instructions()` can later fail startup

src/reachy_mini_conversation_app/static/js/views/home.js
```diff
@@ -193,9 +198,31 @@ export async function mountHomeView({ outlet, signal, navigate }) {
...
+    const ok = await confirmDialog({
+      title: "Delete personality?",
+      message: `"${prettifyProfileName(name)}" will be permanently removed.`,
+      confirmLabel: "Delete",
+      danger: true,
+    });
+    if (!ok || signal.aborted) return;
+    status.classList.remove("is-warning", "is-error");
+    try {
+      await deletePersonality(name);
```
alozowski, line 212: Tiny API contract issue: this removes the card for any 2xx response, even when the backend returns `{ ok: false }`. Could we check the returned ok value or make the route return a non-2xx for refused/missing deletes? That matches the rest of the UI’s error handling through `describeError()`

src/reachy_mini_conversation_app/static/js/components/confirm-dialog.js
```diff
@@ -0,0 +1,49 @@
...
+    );
+    overlay.appendChild(dialog);
+    document.body.appendChild(overlay);
+
+    function close(value) {
+      window.removeEventListener("keydown", onKeydown);
+      overlay.remove();
+      resolve(value);
+    }
+    function onKeydown(event) {
```
alozowski, line 37: Small accessibility bug: global Enter confirms even if focus is on Cancel. Could we only handle Escape here and let the focused button handle Enter/click normally? That keeps this modal consistent with browser dialog expectations

src/reachy_mini_conversation_app/personality.py
```diff
@@ -115,6 +115,24 @@ def available_tools_for(selected: str) -> List[str]:
     return sorted(set(shared + local))
 
 
+def delete_personality(name: str) -> bool:
```
alozowski, line 118: Let's add a regression coverage for this behavior? A delete feature touches persisted user data, so we can add tests for deleting a `user_personalities/...` profile and refusing built-in/outside profiles, plus the startup/current-profile guard from the route

alozowski (approved):
That's a nice feature! I left some comments that are worth addressing, but I don't want to block this PR, so I'm approving in advance

PR #442, opened 2026-06-30
Fall back to default instructions for incomplete profiles

andimarafioti:
## Summary
Missing or empty `instructions.txt` files in a selected profile used to call `sys.exit(1)` during session startup. This makes the prompt loader warn and fall back to `profiles/default/instructions.txt` so an incomplete profile does not tear down the conversation app.

The PR also tightens the runtime personality error path now that the prompt loader no longer exits, adds a regression test, and documents the fallback behavior.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [ ] Console mode (default)
- [ ] Web UI (`--ui`)
</details>

<details>

</details>

## Notes
Validated locally with:

- `.venv/bin/pytest tests/ -v` (266 passed)
- `.venv/bin/mypy --pretty --show-error-codes`
- `.venv/bin/ruff check src/reachy_mini_conversation_app/prompts.py src/reachy_mini_conversation_app/base_realtime.py tests/test_profile_paths.py`
- `.venv/bin/ruff format --check src/reachy_mini_conversation_app/prompts.py src/reachy_mini_conversation_app/base_realtime.py tests/test_profile_paths.py`

`ruff check .` is currently blocked by unrelated local-only files outside this PR (`talk_to_backend_no_robot.py` is untracked and fails public-docstring rules).

### DIFF
--- a/src/reachy_mini_conversation_app/base_realtime.py
+++ b/src/reachy_mini_conversation_app/base_realtime.py
@@ -305,7 +305,7 @@ class BaseRealtimeHandler(ConversationHandler, ABC):
             try:
                 instructions = self._get_session_instructions()
                 voice = self.get_current_voice()
-            except BaseException as e:  # catch SystemExit from prompt loader without crashing
+            except Exception as e:
                 logger.error("Failed to resolve personality content: %s", e)
                 return f"Failed to apply personality: {e}"
 
--- a/src/reachy_mini_conversation_app/prompts.py
+++ b/src/reachy_mini_conversation_app/prompts.py
@@ -1,4 +1,3 @@
-import sys
 import logging
 from pathlib import Path
 
@@ -24,6 +23,22 @@ def _default_instructions_file() -> Path:
     return DEFAULT_PROFILES_DIRECTORY / DEFAULT_PROFILE_NAME / INSTRUCTIONS_FILENAME
 
 
+def _read_instructions_file(instructions_file: Path, profile_name: str) -> str | None:
+    try:
+        if not instructions_file.exists():
+            logger.warning("Profile '%s' has no %s", profile_name, INSTRUCTIONS_FILENAME)
+            return None
+        instructions = instructions_file.read_text(encoding="utf-8").strip()
+    except (OSError, UnicodeError) as e:
+        logger.warning("Failed to load instructions from profile '%s': %s", profile_name, e)
+        return None
+
+    if not instructions:
+        logger.warning("Profile '%s' has empty %s", profile_name, INSTRUCTIONS_FILENAME)
+        return None
+    return instructions
+
+
 def get_session_instructions(instance_path: str | Path | None = None) -> str:
     """Get session instructions, loading from REACHY_MINI_CUSTOM_PROFILE if set."""
     profile = config.REACHY_MINI_CUSTOM_PROFILE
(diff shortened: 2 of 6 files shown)


src/reachy_mini_conversation_app/base_realtime.py
```diff
@@ -305,7 +305,7 @@ async def apply_personality(self, profile: str | None) -> str:
             try:
                 instructions = self._get_session_instructions()
                 voice = self.get_current_voice()
-            except BaseException as e:  # catch SystemExit from prompt loader without crashing
+            except Exception as e:
```
alozowski, line 308: Since we're changing `base_realtime.py` like that, `src/reachy_mini_conversation_app/gemini_live.py` L250 needs the same change:
```
  try:
      _ = get_session_instructions(self.instance_path)
      _ = get_session_voice()
  except Exception as e:          # was: except BaseException
      logger.error("Failed to resolve personality content: %s", e)
      return f"Failed to apply personality: {e}"
```
I know about the backends, but it's a small change

alozowski: The same way with `src/reachy_mini_conversation_app/console.py`, L529:
```
```python
            set_custom_profile(profile)
            try:
                get_session_instructions()
                get_session_voice(default=get_default_voice_for_backend(get_backend_choice()))
            except Exception:               # L529, was: except BaseException
                set_custom_profile(previous_profile)
                raise
        except Exception as e:
            logger.error("Error applying personality '%s': %s", profile, e)
            return f"Failed to apply personality: {e}"
        except Exception as e:              # L535, was: except BaseException
            logger.error("Failed to resolve personality content: %s", e)
            return f"Failed to apply personality: {e}"
```

Note: with both catches narrowed to `Exception`, the second one (L535) becomes
unreachable, because L532 already handles every `Exception` in this case. So the cleaner end state is to
drop the L535 block entirely rather than narrow it. Please, take a look there

src/reachy_mini_conversation_app/prompts.py
```diff
@@ -42,21 +57,23 @@ def get_session_instructions(instance_path: str | Path | None = None) -> str:
...
-                return instructions
-            logger.error("Profile '%s' has empty %s", profile_name, INSTRUCTIONS_FILENAME)
-            sys.exit(1)
-        logger.error("Profile '%s' has no %s", profile_name, INSTRUCTIONS_FILENAME)
-        sys.exit(1)
-    except Exception as e:
-        logger.error("Failed to load instructions from profile '%s': %s", profile_name, e)
-        sys.exit(1)
+    instructions = _read_instructions_file(instructions_file, profile_name)
+    if instructions is None and profile:
```
alozowski: if someone sets `REACHY_MINI_CUSTOM_PROFILE=default` literally and it's broken, the log reads "falling back to default because profile 'default' is incomplete" and the default file is read twice – a cosmetic and rare thing that I noticed

alozowski (changes requested):
It's a good fix, let's merge after one consistency change

src/reachy_mini_conversation_app/base_realtime.py
andimarafioti: Done in 4af5fa62. I narrowed the Gemini catch to `Exception`, narrowed the console rollback catch, and removed the now-unreachable console `BaseException` block.

src/reachy_mini_conversation_app/prompts.py
andimarafioti: Good catch, fixed in 4af5fa62. The explicit `default` profile now skips the fallback-to-default path, and there is a regression test for the self-fallback log case.

alozowski (approved):
Let's go!

Issue #315, opened 2026-04-22
Require fresh camera capture for every visual question


alozowski:
Every new visual question should call the camera unless the user explicitly refers to the last picture.

wsrast:
Some evidence on the mechanism here, since this may not be fixable by
prompting alone.

On a Wireless unit (conversation app 1.0.0, reachy_mini 1.10.0rc2, daemon
1.9.0) the camera *is* being called for each visual question, and the capture
is fresh — but the answer is still one turn stale:

```
18:52:01.525 huggingface_realtime:862 Tool call received — tool_name='camera'
18:52:01.525 background_tool_manager:163 Started background tool: camera
18:52:01.550 tools.camera:45 Tool call: camera question=...
18:52:01.586 huggingface_realtime:583 Tool 'camera' executed successfully
18:52:01.749 huggingface_realtime:675 Added camera image to conversation
jpeg_bytes=330015
```

`executed successfully` is logged 163 ms *before* the image is appended to the
conversation, so the model can generate its response before the image is
visible to it. The image is then in context for the *following* turn.

The capture itself is fine — ~220 ms end to end, and `jpeg_bytes` differs
between calls, so these are genuinely distinct frames.

`huggingface_realtime.py:881` dispatches every tool call through
`tool_manager.start_tool(...)` with no per-tool branching (the only check on
`tool_name` is that it's a string), so this looks like an intended consequence
of background dispatch rather than a defect in the camera tool.

Two implications for this issue:

1. Instructing the model to always call the camera won't be sufficient — the
   freshly captured image still lands a turn late.
2. The stale image left in context by the late append is plausibly part of
   *why* the model skips re-calling the camera, so the two behaviours may be
   the same root cause.

Happy to test a fix.

Issue #314, opened 2026-04-22
Default bare camera requests to "describe what you see"


alozowski:
Bare picture requests should trigger the camera instead of asking for clarification. For example, a bare "take a picture" produced clarification questions instead of calling camera.

PR #454, opened 2026-07-12
feat: improve camera tool description for better LLM triggering

SunixLiu:
## What changed

  Expanded the `camera` tool description and parameter schema with concrete use cases so the LLM knows when to call it.

  ### Before
  - Single-line description: "Take a picture with the camera and ask a question about it."
  - Minimal parameter description with no examples

  ### After
  - Multi-scene description covering: looking at objects, checking appearance, describing scenes, visual opinions
  - Parameter `question` now includes examples ("what is the user holding", "describe the users outfit", etc.)
  - Docstring corrected to accurately describe the return value (base64-encoded JPEG)

  ## Why

  The camera tool was rarely triggered by the LLM because the description was too vague. With concrete use cases, the LLM can
  better match user intent to the right tool.

  ## Notes

  This is a temporary improvement for the current `main` branch. The upstream `fix/201-distinct-vision-tools` branch is working on
  splitting camera into two tools (`describe_camera_image` and `see_image_through_camera`), and this description enhancement can
  be ported there once it merges.


alozowski:
Hello @SunixLiu, 

Thanks for this, it's a good change! Vague tool descriptions are exactly why the camera rarely fired, and this lines up with what we're tracking in #314. Three small things before we merge:
- a small type: "describe the users outfit" should be "describe the user's outfit" (the model reads these, so worth getting right)
- could you add one more sentence to the description: "The camera is live, each call captures the current moment." That nudges the model to take a fresh picture for new visual questions instead of reusing an old one (see #315).
- one more behavioural nudge, per OpenAI's realtime prompting guidance on tool descriptions: "If the user asks you to look without saying at what, do not ask for clarification, call this tool and describe what you see." That targets the exact failure where the model interrogates the user instead of just looking.

Happy to merge once those are in!

SunixLiu:
Thanks for the feedback! Addressed all three points:

1. Fixed typo: "users outfit" → "user's outfit"
2. Added: "The camera is live, each call captures the current moment."
3. Added: "If the user asks you to look without saying at what, do not ask for clarification, call this tool and describe
what you see."

All checks pass (ruff, mypy strict, pytest).

alozowski:
Thank you for addressing the comments! One last mechanical thing: your editor re-saved the file as UTF-8 with BOM and Windows (CRLF) line endings, so github now shows the whole file as a binary change and we can't see a reviewable diff. Could you re-save it as UTF-8 without BOM with LF line endings? In vs code that's the "UTF-8" and "LF" selectors in the status bar. After that we'll merge

SunixLiu:
> Thank you for addressing the comments! One last mechanical thing: your editor re-saved the file as UTF-8 with BOM and Windows (CRLF) line endings, so github now shows the whole file as a binary change and we can't see a reviewable diff. Could you re-save it as UTF-8 without BOM with LF line endings? In vs code that's the "UTF-8" and "LF" selectors in the status bar. After that we'll merge

This is done, please let me know  if it's Ok now.

alozowski (approved):
Verified the file is now clean UTF-8/LF and the diff is exactly the agreed changes, merging. Thanks @SunixLiu!

PR #510, opened 2026-07-23
Enable motors before waking from sleep

andimarafioti:
## Summary
Enable motor torque before running the SDK wake-up movement when startup detects that Reachy is in the sleep pose.

The desktop app's daemon leaves the robot limp after stopping an app in the sleep pose. The conversation app previously called `wake_up()` without re-enabling the motors, so its motion targets had no effect after a stop/start cycle. A regression test now verifies that motors are enabled before `wake_up()` and remain untouched when Reachy is already awake.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [ ] CI is green (lint, types, tests — runs on Linux, macOS & Windows)
- [x] Code is clear (types, docs, comments where needed)
- [ ] `.env.example` updated (if new config vars were added)
- [ ] Installation from the desktop app tested (if applicable)

<details>
<summary><strong>UI / run modes</strong> — expand if your PR touches these</summary>

- [ ] Console mode (default)
- [ ] Web UI (`--ui`)
</details>

<details>
<summary><strong>Vision / motion</strong> — expand if your PR touches these</summary>

- [ ] Camera pipeline (with/without `--no-camera`)
- [ ] Movement manager (dances, emotions, head motion)
- [ ] Profiles or custom tools
</details>

## Notes
Validated from an isolated worktree with Ruff, formatting, strict mypy, and all 231 tests. The wheel-path test required overriding a timed-out local Hugging Face package index with `UV_DEFAULT_INDEX=https://pypi.org/simple`.

PR #512, opened 2026-07-25
Standby mode: local wake word instead of app exit on go_to_sleep (opt-in)

saintpete:
## Motivation

For an always-on (especially Wireless) Reachy Mini, `go_to_sleep` currently stops the app, so waking means a full app restart (~15–25 s). This PR adds an **opt-in standby mode** aimed at privacy-conscious always-on use:

- "Go to sleep" closes the realtime backend connection (zero audio leaves the robot) but keeps the app alive.
- Mic frames are routed only to an on-device [openWakeWord](https://github.com/dscripka/openWakeWord) model (default phrase: "hey Jarvis").
- On detection, the robot physically wakes and the backend reconnects — measured **under 3 seconds** voice-to-ready on a Wireless unit, vs the current full restart.

## Design

- `STANDBY_ON_SLEEP=1` env flag (default **off** — stock behavior unchanged), plus `WAKE_WORD_MODEL` / `WAKE_WORD_THRESHOLD`.
- New `standby.py`: streaming `WakeWordDetector` (any input rate → mono 16 kHz int16 chunks, ONNX inference only).
- `LocalStream`: `enter_standby()` / `exit_standby()` with startup-loop gating so the retry loop can't silently reconnect while asleep; `conversation.wake` RPC method; `in_standby` in status.
- `main.py`: the `go_to_sleep` tool branches into standby; the inactivity timer re-arms across sleep/wake cycles; a wake hook re-enables motors/wobbling (builds on #510).
- **Fail-safe**: if the detector or handler factory is unavailable, standby refuses and the existing stop-the-app path runs — the robot is never left unwakeable.
- `openwakeword` is an optional `[wakeword]` extra.

## Testing

- Unit tests for the detector (buffering, thresholds, resampling, missing-dependency fallback).
- End-to-end on a physical Reachy Mini Wireless (RPi CM4, Python 3.12) via a v0.10.0 backport of this branch (`standby-v0.10.0` on the fork, since the robot ships SDK 1.9.0): inactivity → standby (backend closed, wake word armed) → wake → reconnected in <3 s. Detector validated on-device with synthesized positive/negative phrases.
- Marked **draft** because this `main`-based branch itself hasn't run on hardware with SDK 1.10 yet.

## Caveat worth discussing

`openwakeword>=0.6.0` declares `tflite-runtime` on Linux, which has **no Python 3.12 wheels** — on the robot we installed it `--no-deps` (the ONNX path needs nothing from tflite). Happy to adapt if you'd prefer vendoring the ONNX inference, or wiring in pollen-robotics/wake-word-detector instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,10 @@ dependencies = [
     "mcp>=1.27.1",
 ]
 
+[project.optional-dependencies]
+# Local wake word for standby mode (STANDBY_ON_SLEEP=1)
+wakeword = ["openwakeword>=0.6.0"]
+
 [dependency-groups]
 dev = [
   "allure-pytest==2.16.0",
--- a/src/reachy_mini_conversation_app/config.py
+++ b/src/reachy_mini_conversation_app/config.py
@@ -120,8 +120,26 @@ def _env_flag(name: str, default: bool = False) -> bool:
     return default
 
 
+def _env_float(name: str, default: float) -> float:
+    """Parse a float environment value, falling back to ``default`` on errors."""
+    raw = (os.getenv(name) or "").strip()
+    if not raw:
+        return default
+    try:
+        return float(raw)
+    except ValueError:
+        logger.warning("Invalid float value for %s=%r, using default=%s", name, raw, default)
+        return default
+
+
 APP_TIMEOUT_MINUTES_ENV = "REACHY_MINI_APP_TIMEOUT_MINUTES"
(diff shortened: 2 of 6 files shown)


saintpete:
Superseded by #514 — the same standby/local-wake-word feature, re-extracted onto current main (post JSON-RPC refactor), now with detector unit tests and on-device measurements (0.00 false accepts/hr over 75 min of real room audio, 91% recall). Closing this draft in favor of that one.

PR #513, opened 2026-07-30
perf(startup): warm openai+scipy imports during robot init; pre-import the openai realtime tree

RemiFabre:
Two commits that move import cost off the app-start critical path: total work is unchanged, but it runs during phases where the main thread is just waiting (robot connect, the ~2 s wake move, media init), so it costs ~zero wall-clock.

1. **Warmup thread**: a background thread imports openai + scipy while the robot initializes. Pairs with the SDK import diet (pollen-robotics/reachy_mini#1308), which makes those imports deferrable in the first place — you can't warm what eager imports already loaded before `main()` ran.
2. **openai realtime pre-import**: the "2.2 s websocket connect" was profiled at ~70 % CPU — `AsyncOpenAI.realtime` is a `cached_property` whose module tree (~1.6 s on the CM4) loaded lazily *inside* the connect phase, masquerading as network time. Two import lines in the warmup thread fix it (true network cost of the connect: ~470 ms).

**Measured** (2026-07-30, wireless robot with #1308 installed, standard warm protocol: fresh daemon → sleep pose → start-app POST → "usable" = max(session updated, audio config applied); 3 runs each, first run after each install discarded):

| app start → usable | run 1 | run 2 | run 3 | mean |
|---|---|---|---|---|
| current HF release | 13.78 | 13.66 | 13.96 | **13.8 s** |
| this branch | 10.03 | 10.05 | 10.28 | **10.12 s** |

**−3.7 s (−27 %)** on the wake-later path — the flow users actually experience (robot boots asleep, app starts on wake). Cold power-on → talking is *not* claimed by this PR: during boot all cores are busy, so there are no idle windows to hide imports in.

**Risk**: ~nil — imports only, no reordering of audio/motion/session logic. If the warmup thread dies, behavior is exactly today's (imports happen at first use). 294 non-audio tests pass; ruff/mypy clean.

### DIFF
--- a/src/reachy_mini_conversation_app/main.py
+++ b/src/reachy_mini_conversation_app/main.py
@@ -26,6 +26,34 @@ if TYPE_CHECKING:
     from reachy_mini_conversation_app.console import LocalStream
 
 
+_import_warmup_thread: threading.Thread | None = None
+
+
+def start_import_warmup() -> threading.Thread:
+    """Preload heavy modules in the background while the robot connects.
+
+    openai (~1.5s) and scipy (~1s on the robot) are needed by the realtime
+    handler and the movement manager, but not before the robot / media
+    pipelines are up. Importing them on a side thread overlaps that cost with
+    robot initialization instead of paying it serially. Idempotent; errors are
+    deferred to the real import site, which reports them with full context.
+    """
+    global _import_warmup_thread
+    if _import_warmup_thread is None:
+
+        def _preload() -> None:
+            try:
+                import scipy.spatial.transform  # noqa: F401
+
+                import reachy_mini_conversation_app.huggingface_realtime  # noqa: F401
+            except Exception:
+                logging.getLogger(__name__).debug("Import warmup failed", exc_info=True)
+
+        _import_warmup_thread = threading.Thread(target=_preload, name="import-warmup", daemon=True)
+        _import_warmup_thread.start()
+    return _import_warmup_thread
+
+
 def _start_inactivity_timeout_thread(
     timeout_minutes: float,
     stream_manager: LocalStream,
@@ -84,6 +112,7 @@ def run(
     instance_path: Optional[str] = None,
 ) -> None:
     """Run the Reachy Mini conversation app."""
+    start_import_warmup()
     # Putting these dependencies here makes the dashboard faster to load when the conversation app is installed
     from reachy_mini_conversation_app.moves import MovementManager
     from reachy_mini_conversation_app.config import (
@@ -351,6 +380,11 @@ class ReachyMiniConversationApp(ReachyMiniApp):  # type: ignore[misc]
 
     custom_app_url = "http://0.0.0.0:7860/"
     dont_start_webserver = False
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Start preloading heavy imports before the robot/media init in wrapped_run."""
+        start_import_warmup()
+        super().__init__(*args, **kwargs)
 
     def run(self, reachy_mini: ReachyMini, stop_event: threading.Event) -> None:
         """Run the Reachy Mini conversation app."""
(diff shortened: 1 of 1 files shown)


RemiFabre:
### How to test on a robot

*(Executed end-to-end on a wireless unit 2026-07-30; the numbers in the description are those runs.)*

Best measured together with the SDK import diet (pollen-robotics/reachy_mini#1308) installed — the warmup thread warms the imports that PR defers:

    JOB=$(curl -s -X POST "http://localhost:8000/update/start-from-ref?git_ref=pr/import-diet" | sed 's/.*"job_id":"\([^"]*\)".*/\1/')
    # wait until install-source shows pr/import-diet (the daemon restarts itself at the end)

**1. Baseline** (HF-release app), repeat 3×; discard the first run after any install (pyc):

    sudo systemctl restart reachy-mini-daemon && sleep 15
    curl -s -X POST localhost:8000/api/move/play/goto_sleep >/dev/null && sleep 8
    T0=$(awk '{print $1}' /proc/uptime)
    curl -s -X POST localhost:8000/api/apps/start-app/reachy_mini_conversation_app >/dev/null
    sleep 25; journalctl -b -o short-monotonic --since "-2 min" | grep -E "Realtime session updated|audio startup config" | head -2
    # usable = max(marker timestamps) - T0.  Expected: ~13.8 s
    curl -s -X POST localhost:8000/api/apps/stop-current-app >/dev/null

**2. Install this branch into the apps venv** (dashboard updates refuse non-HF sources, so use pip directly):

    /venvs/apps_venv/bin/pip install --force-reinstall --no-deps "git+https://github.com/pollen-robotics/reachy_mini_conversation_app@pr/warmup-preimport"

**3. Re-run step 1.** Expected: **~10.1 s** (−3.7 s). Listen once: the greeting must sound unchanged — this PR moves imports only, it does not reorder audio.

**4. Restore**: reinstall the app from the dashboard (HF), and `start-from-ref?git_ref=main` for the SDK.

Pitfalls: ~15 app starts in a day can trip the HF allocator's rate limit (429 on POST /session — the app then retries forever and never becomes usable; not this PR). And space runs ~20 s apart so the previous session is fully torn down.

FabienDanieau (approved):
Tested on a wireless unit (apps venv on SDK 1.10.0), same protocol as above, 3 runs each, first run after each install discarded:

| app start → usable | run 1 | run 2 | run 3 | mean |
|---|---|---|---|---|
| installed 1.0.1 | 11.59 | 11.57 | 11.57 | **11.58 s** |
| this branch | 8.82 | 9.62 | 9.55 | **9.33 s** |

**-2.25 s (-19 %)**, no 429s, no tracebacks, no `Import warmup failed`. My baseline is 11.58 s rather than 13.8 s (different robot state), so the absolute delta is smaller than yours, but it sits well outside the noise: baseline spread was ±0.02 s.

Measured the mechanism standalone on the same robot: importing `openai.resources.realtime` + `websockets.asyncio.client` on the main thread costs 3.3-3.5 s, and with a 2 s background overlap only 1.33-1.39 s is left. So the tree is bigger than the ~1.6 s in the description, and the win is bounded by the size of the idle window rather than by the import cost. Roughly 1.3 s is still unabsorbed.

Also checked: the daemon's app listing reads `custom_app_url` off the class (`ep.load()`), so importing the module does not start the thread and the dashboard stays fast; and the package has no runtime import cycle (`core_tools` ↔ `background_tool_manager` is TYPE_CHECKING-only), so two threads importing concurrently can't hit the module-lock deadlock path.

Two non-blocking nits:
- Title and description still mention scipy, dropped in 4a597ac.
- A failed warmup is invisible: the `.debug()` fires before `setup_logger()` has configured logging, so if `openai.resources.realtime` ever moves, the PR silently stops working and the 2 s comes back unnoticed. A three-line test importing the three module paths would catch that on the next openai bump.

One thing I could not check remotely: whether the CPU-bound import on a side thread adds any jitter to the wake move. Worth one look before merge.

PR #514, opened 2026-08-02
Standby mode: "go to sleep" arms a local wake word instead of exiting

saintpete:
## What

Opt-in standby for the conversation app: with `STANDBY_ON_SLEEP=1`, going to sleep closes the realtime backend connection and routes mic frames **only** to a local [openWakeWord](https://github.com/dscripka/openWakeWord) detector. Audio is scored on-device in 80 ms windows — never buffered beyond the analysis window, never stored, never sent anywhere. On detection, the robot runs its physical wake-up move, then reconnects the backend fresh.

Default is **off**: without the env var there is zero behavior change.

## Why

A sleeping robot that keeps its realtime session open is paying for silence; one that exits entirely can't be woken by voice. Standby is both the privacy guarantee (no audio leaves the robot while asleep) and the cost story (standby minutes are free) — the robot stays *available* without staying *connected*.

## How it works

- `STANDBY_ON_SLEEP=1` arms the feature; sleep intent enters standby instead of exiting.
- `WAKE_WORD_MODEL` — a pretrained openWakeWord name (default `hey_jarvis_v0.1`) **or a path to a custom `.onnx`**, so owners can train their robot's actual name.
- `WAKE_WORD_THRESHOLD` — default 0.5.
- `standby.wake` RPC method wakes it remotely; `conversation.status` exposes `in_standby`.
- openwakeword ships as an optional extra (`pip install .[wakeword]`), ONNX inference path.

Safety-of-behavior details:
- **Refuses rather than degrades**: if the detector can't be built or the handler can't be rebuilt later, entering an unwakeable standby would strand the robot — so it declines and stays connected.
- Mic frames stop reaching the backend the instant sleep is requested, before the (possibly slow) first detector build.
- The backend stays gated until the wake move completes, so the model can't speak into its own shell mid-tuck (muffled self-audio otherwise reads as barge-in).

## Testing

- `tests/test_standby.py` covers the detector wrapper (chunking, resampling, threshold, reset) with a fake model — no openwakeword needed in CI.
- Field-tested for several days as the daily driver on a Reachy Mini (as a backport of this branch), including a custom-trained three-phrase wake word that measured **0.00 false accepts/hour over 75 minutes of real living-room audio at 91% recall** on the device.
- For owners who want their robot's own name: a companion toolkit for training + measuring custom models on current Colab lives at https://github.com/saintpete/openwakeword-colab-toolkit (public domain).

One install note: on Python 3.12, openwakeword's `tflite-runtime` dependency has no wheels, so a plain `pip install openwakeword` can backtrack to an old version — installing the extra with `--no-deps` plus `onnxruntime` is the reliable pattern on 3.12 (the detector here only uses the ONNX path). Happy to add that to docs wherever you'd like it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,10 @@ dependencies = [
     "mcp>=1.27.1",
 ]
 
+[project.optional-dependencies]
+# Local wake word for standby mode (STANDBY_ON_SLEEP=1)
+wakeword = ["openwakeword>=0.6.0"]
+
 [dependency-groups]
 dev = [
   "allure-pytest==2.16.0",
--- a/src/reachy_mini_conversation_app/config.py
+++ b/src/reachy_mini_conversation_app/config.py
@@ -121,8 +121,26 @@ def _env_flag(name: str, default: bool = False) -> bool:
     return default
 
 
+def _env_float(name: str, default: float) -> float:
+    """Parse a float environment value, falling back to ``default`` on errors."""
+    raw = (os.getenv(name) or "").strip()
+    if not raw:
+        return default
+    try:
+        return float(raw)
+    except ValueError:
+        logger.warning("Invalid float value for %s=%r, using default=%s", name, raw, default)
+        return default
+
+
 APP_TIMEOUT_MINUTES_ENV = "REACHY_MINI_APP_TIMEOUT_MINUTES"
 DEFAULT_APP_TIMEOUT_MINUTES = 1440.0
+
+STANDBY_ON_SLEEP_ENV = "STANDBY_ON_SLEEP"
+WAKE_WORD_MODEL_ENV = "WAKE_WORD_MODEL"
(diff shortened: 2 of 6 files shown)


FabienDanieau:
Hey, thanks for the PR. I've started to work on something similar on [the daemon side](https://github.com/pollen-robotics/reachy_mini/issues/1253). We feel it makes more sense to have it there since the robot could wake up and then launch any start up app (not necessarily the conv app). Also I've chosen nanowakeword that uses onnx which is already a dependency of reachy mini. Ideally we'd like only one inference engine (shared with the head tracking). Let's discuss !

FabienDanieau:
cc @andimarafioti , the wake up word procedure will have to be combined with the wake up triggered by the antennas

saintpete:
@FabienDanieau absolutely. I suspected this code wouldn't get merged so much as share the process I went through to achieve the outcome, and let you take anything useful from it.

PR #484, opened 2026-07-17
Improve profile tools UI

alozowski:
## Summary
Replaces `.txt` profile files with strict `profile.md` documents and shared tools. Adds per-profile built-in, external, and MCP tool access with the complete "Profiles" and "Tools" UI. Updates runtime loading, persistence, packaging, documentation, and essential regression coverage.

Closes https://github.com/pollen-robotics/reachy_mini_conversation_app/issues/449

## Category
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [x] CI is green (lint, types, tests — runs on Linux, macOS & Windows)
- [x] Code is clear (types, docs, comments where needed)
- [ ] `.env.example` updated (if new config vars were added)
- [x] Installation from the desktop app tested (if applicable)

<details>
<summary><strong>UI / run modes</strong> — expand if your PR touches these</summary>

- [ ] Console mode (default)
- [ ] Web UI (`--ui`)
</details>

<details>
<summary><strong>Vision / motion</strong> — expand if your PR touches these</summary>

- [ ] Camera pipeline (with/without `--no-camera`)
- [ ] Movement manager (dances, emotions, head motion)
- [ ] Profiles or custom tools
</details>

PR #530, opened 2026-08-13
fix(profiles): keep profiles authored before the profile.md format working

### LINKED ISSUE #529: App exits with code 1 when the selected profile can no longer be read
## Symptom

The app refuses to start and dies immediately. From the user's side it looks like "the conversation app crashes as soon as I start it / click the microphone button". The daemon reports the app in `error` state:

```
ERROR reachy_mini_conversation_app.tools.core_tools:302 | Failed to read tools for profile 'user_personalities/famille_vacances': Profile 'user_personalities/famille_vacances' has no profile.md at /venvs/apps_venv/lib/python3.12/site-packages/reachy_mini_conversation_app/user_personalities/famille_vacances
ERROR reachy_mini_conversation_app.utils:283 | Failed to initialize tools: Failed to read tools for profile 'user_personalities/famille_vacances'
ERROR App reachy_mini_conversation_app exited with code 1.
```

Observed on a Reachy Mini Wireless running daemon `v1.10.0rc5`, conversation app installed from the app store.

## Root cause

The selected profile is persisted, so it outlives the app version that created it. It can stop being readable later:

- the profile predates the `profile.md` format and its directory only carries the old `greeting.txt` / `instructions.txt` (this is what happened here: files dated well before the reinstall, no `profile.md` anywhere in the package)
- the directory is partly removed by an uninstall/reinstall cycle

`core_tools._read_profile_tool_names` re-raises as `RuntimeError`, and `main.py` turns any `initialize_tools` failure into `sys.exit(1)`.

## Why this is worse than a normal startup error

It is unrecoverable from the product. The profile can only be changed from the settings UI, and the settings UI is served by the app that refuses to start. The only way out is to hand-edit files inside `site-packages` over SSH, which is not something a user can do.

Note the app is already resilient about this **everywhere else**: `prompts.py` catches `FileNotFoundError` / `ProfileFormatError` and degrades to the packaged default for instructions (`get_session_instructions`), voice (`get_session_voice`) and greeting (`get_session_greeting_prompt`). The tool registry is the only path that treats it as fatal, so the app dies for the one part of the profile that has a safe default.

## Reproduce

```bash
mkdir -p <profiles_root>/legacy_profile
echo "tu es un robot" > <profiles_root>/legacy_profile/instructions.txt   # no profile.md
# select "legacy_profile" as the active profile, then start the app
```

The app exits 1 on every subsequent start.

## Suggested fix

Degrade to the packaged default profile and log the abandoned profile loudly, rather than exiting. Keep `initialize_tools` raising so library callers still get a catchable failure (`test_missing_profile_raises_runtime_error_instead_of_exiting`) and put the recovery policy in the app layer.

## Related, not fixed here

- **User profiles live inside `site-packages`** (`.../site-packages/reachy_mini_conversation_app/user_personalities/`). Anything a user authors there is at the mercy of the next reinstall. Worth moving under the instance/config dir.
- **No migration for legacy profiles.** A profile with `instructions.txt` + `greeting.txt` and no `profile.md` is silently unusable rather than being migrated. Migrating on read would have preserved this user's content instead of dropping them to the default persona.

RemiFabre:
## Summary

The app exits with code 1 on every start when the selected profile predates the `profile.md` format. That is unrecoverable in practice: the profile files and the `startup_settings.json` selection are untracked files inside site-packages, so they survive uninstall/reinstall (pip and uv only remove RECORD-listed files; verified empirically), and the profile can only be changed from a settings UI the app has to be running to serve. Closes #529.

#484 converted the bundled profiles to `profile.md` but nothing converted user profiles under `user_personalities/`, which we cannot enumerate at release time.

Design, reworked after review:

1. **Migrate at startup.** `migrate_legacy_profiles` runs once at app startup, scoped to the `user_personalities` root, and writes a `profile.md` for every sidecar-file profile (`instructions.txt`, `tools.txt`, `voice.txt`, `greeting.txt`). Sidecar files stay in place and `profile.md` wins on later reads, so the conversion is additive and idempotent. A profile with no `tools.txt` inherits the bundled default's tool list, which is what pre-#484 builds did and therefore what those profiles were authored against. The canonical reader and the profile listing stay strict about `profile.md`: no second supported format, no warn-once machinery.
2. **Never crash on a broken profile.** When the selected profile is unreadable for a reason the migration cannot rescue (malformed `profile.md`, half-removed directory), fall back to the default profile instead of `sys.exit(1)`. `initialize_tools` keeps raising, so library callers still get a catchable failure.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist

<details>

- [x] Console mode (default)
- [ ] Web UI (`--ui`)
</details>

<details>

</details>

## Notes

319 tests, `ruff` and `mypy` clean locally (macOS). No new config vars.

Written with AI assistance; commits carry the `Assisted-by` trailer.

<details>
<summary><strong>Annex</strong> — hardware verification (Reachy Mini Wireless, daemon <code>v1.10.0rc5</code>)</summary>

Protocol: restore the pristine released 1.0.0 files into the venv, create a legacy profile (`instructions.txt` + `greeting.txt`, no `profile.md`), select it, start via the daemon API.

**Before (pristine 1.0.0)**

```
ERROR core_tools:302 | Failed to read tools for profile 'user_personalities/repro_legacy':
      has no profile.md at .../user_personalities/repro_legacy
ERROR utils:283      | Failed to initialize tools
ERROR App reachy_mini_conversation_app exited with code 1.
```

**After (this branch patched in, same profile, same selection)**

```
INFO profile_store:189       | Migrated legacy profile 'repro_legacy' to profile.md.
INFO core_tools:299          | Loading tools for profile: user_personalities/repro_legacy
INFO huggingface_realtime:712| Realtime session initialized with
     profile='user_personalities/repro_legacy' voice='Aiden'
```

The written `profile.md` carries the original instructions and greeting, and the inherited default tool list (`sweep_look`, `search_web`, `get_time`, `get_weather` included), reproducing the pre-#484 behaviour rather than approximating it.

Fallback path, same robot, `profile.md` then overwritten with garbage:

```
ERROR utils:62 | Profile 'user_personalities/repro_legacy' could not be loaded (...);
      starting on the 'default' profile instead.
INFO  core_tools:299 | Loading tools for profile: default
```

App state after both runs: `running`, no exit.

</details>

<details>
<summary><strong>Not addressed here</strong></summary>

User profiles and `startup_settings.json` live inside site-packages as files pip does not track. That is exactly why this bug survives uninstall/reinstall cycles, and also why those files would be lost if the venv were ever rebuilt. Moving them under the instance directory is the deeper issue behind #529 and wants a separate change.

</details>


src/reachy_mini_conversation_app/profile_store.py
```diff
@@ -123,10 +131,83 @@ def _parse_profile_document(source_path: Path) -> ProfileDefinition:
...
+        greeting=_read_optional_legacy_text(profile_directory / LEGACY_GREETING_FILENAME),
+        hidden=False,
+    )
+
+
 def read_profile_from_directory(profile_name: str, profile_directory: Path) -> ProfileDefinition:
     """Read a profile document from its directory."""
     profile_path = profile_directory / PROFILE_FILENAME
     if not profile_path.is_file():
+        legacy_profile = read_legacy_profile_from_directory(profile_name, profile_directory)
```
alozowski: Accepting txt here makes the legacy layout a second supported format for every profile root. Could we keep the canonical reader strict and invoke the compatibility adapter only for old `user_personalities`?

src/reachy_mini_conversation_app/profile_store.py
```diff
@@ -148,8 +229,14 @@ def read_packaged_default_profile() -> ProfileDefinition:
...
 
 def profile_directory_has_definition(profile_directory: Path) -> bool:
-    """Return whether a directory contains a profile definition."""
-    return (profile_directory / PROFILE_FILENAME).is_file()
+    """Return whether a directory contains a profile definition.
+
+    Legacy directories count, so a profile that is readable is also selectable
+    rather than working while staying invisible in the profile list.
+    """
+    return (profile_directory / PROFILE_FILENAME).is_file() or (
```
alozowski: I suggest to have `profile.md` as a required profile definition, so could we keep this predicate strict and let only `user_personalities` discovery opt into the temporary txt compatibility path?

src/reachy_mini_conversation_app/profile_store.py
```diff
@@ -123,10 +131,83 @@ def _parse_profile_document(source_path: Path) -> ProfileDefinition:
     )
 
 
+_WARNED_LEGACY_PROFILE_DIRS: set[str] = set()
```
alozowski: A global set, a second lock, and a single-use marker feel heavy if we would like to introduce a temporary compatibility shim (can be discussed, that just my suggestion to keep this compatibility feature temporary) – if you agree, we could use a small cached warning function instead, because a duplicate warning during a simultaneous first read would be an acceptable trade-off here

src/reachy_mini_conversation_app/app_lifecycle.py
```diff
@@ -21,6 +24,54 @@
 _SLEEP_HEAD_ROTATION_TOLERANCE_RAD = 0.35
 
 
+def initialize_tools_with_default_fallback(
+    instance_path: str | Path | None,
+    logger: logging.Logger,
+) -> str | None:
+    """Load the tool registry, degrading to the default profile if need be.
```
alozowski, line 31: Could we reduce the new public and test docstrings to one-line descriptions of current behavior and keep the migration history in the PR body? That will also make the compatibility code easier to remove later (as I mentioned earlier, I suggest to have this code a temp one)

tests/test_profile_load_resilience.py
```diff
@@ -0,0 +1,213 @@
...
+
+    _use_profile(monkeypatch, profiles_root, "locked_profile")
+    monkeypatch.setattr(config_mod, "LOCKED_PROFILE", "locked_profile", raising=False)
+
+    calls: list[object] = []
+    real_initialize_tools = app_lifecycle.initialize_tools
+
+    def _counting(*args: object, **kwargs: object) -> None:
+        calls.append(object())
+        return real_initialize_tools(*args, **kwargs)  # type: ignore[arg-type]
```
alozowski: Could we give `_counting` the same signature as `initialize_tools` and remove this new `type: ignore`? That keeps the test typed without suppressing the mismatch.

tests/test_profile_load_resilience.py
```diff
@@ -0,0 +1,213 @@
...
+
+    monkeypatch.setattr(app_lifecycle, "initialize_tools", _counting)
+
+    with caplog.at_level(logging.ERROR):
+        with pytest.raises(Exception):
```
alozowski: `initialize_tools` wraps this failure as `RuntimeError`, so could we assert that specific type so an unrelated test error cannot satisfy the assertion?

alozowski (changes requested):
I agree that we need this fallback behavior, but I suggest to keep this as a thin migration bridge: `profile.md` should remain the required format, while legacy txt support should exist only at the old `user_personalities` boundary. The current implementation makes txt a supported format across every profile root and adds more permanent machinery than the temporary path needs

alozowski:
@RemiFabre I also opened a related PR in reachy_mini and assigned you as a reviewer, please, check
https://github.com/pollen-robotics/reachy_mini/pull/1357

RemiFabre:
Hello @alozowski,
Thanks for your review. While I agree with most of your comments (and I'll iterate today on this work), there is a high level opinion that I strongly disagree with.
This is the third time that a rework on the format of the profiles breaks compatibility with previous versions and that there is no clean management of the transition/problem. It also creates the type of errors our users will struggle with. Like if it was a matter of uninstall/reinstall it wouldn't be that serious, but these type of errors are persistent between re-installs because of the files created during profile creation. I've had to guide people to SSH into their robot, search for the old file and delete it. They were stuck otherwise.

If we make the transition code temporary, then we're saying to our users: "one day you'll update our software and that day it will stop working for you. You will not know why and you won't know how to fix it." That's an experience I've had so many times with so many pieces of software and that I absolutely hate. I just want the thing to work. And if the thing changed, then guide me through what I need to do to get the thing to work again.

Like this user apparently spent time stuck until it reached for help:
https://discord.com/channels/519098054377340948/1539278587985789020

Also these problems are the ones that require urgent fixes (or an urgent revert), because the risk/reward ratio is huge.

RemiFabre:
I'm going to try a different approach that "self heals" once and for all (we've done that several times in the daemon).
Also I think the most important fix of all is that the default profile runs if anything bad happens (at least users are not stuck)

alozowski:
@RemiFabre I see your point, and I'm thinking about the best solution – it can be either get back to two txt files or keep the support of a txt files as a not temp fallback. I don't have a best solution in mind right now but I'm open to any of your suggestions

alozowski:
> Also I think the most important fix of all is that the default profile runs if anything bad happens

totally yes

RemiFabre:
> @RemiFabre I see your point, and I'm thinking about the best solution – it can be either get back to two txt files or keep the support of a txt files as a not temp fallback. I don't have a best solution in mind right now but I'm open to any of your suggestions

What do you think of the current version?

src/reachy_mini_conversation_app/profile_store.py
```diff
@@ -123,6 +129,68 @@ def _parse_profile_document(source_path: Path) -> ProfileDefinition:
...
+        if (
+            not profile_directory.is_dir()
+            or (profile_directory / PROFILE_FILENAME).is_file()
+            or not (profile_directory / LEGACY_INSTRUCTIONS_FILENAME).is_file()
+        ):
+            continue
+        profile_name = profile_directory.name
+        try:
+            legacy = _read_legacy_profile(profile_name, profile_directory)
+            write_profile(
```
alozowski, line 180: Should we pass `overwrite=False` here? The UI is already available while migration runs, so this could overwrite a profile the user has just saved with the older txt content

alozowski (dismissed):
Thank you for this work! Now the profile versions are coherent

PR #531, opened 2026-08-14
Fix/524 and migrate to mcp v2 sdk

### LINKED ISSUE #524: Conversation app cannot call any remote MCP tool with mcp 2.0.0 - two 1.x-era API usages in mcp_client.py
## Summary

With `mcp` 2.0.0 installed in `apps_venv`, every remote MCP tool call fails.
Two separate 1.x-era API usages in `mcp_client.py` are responsible. Fixing the
first exposes the second.

## Environment

- Robot: Reachy Mini Wireless
- reachy_mini_conversation_app 1.0.0
- reachy_mini 1.10.0rc2 (apps_venv) / daemon 1.9.0
- mcp 2.0.0
- Python 3.12, Debian, kernel 6.12.62+rpt-rpi-v8
- Control app v0.9.33 on macOS

## Bug 1 — three-value unpack of a two-value yield

Any remote tool call fails with:

```
McpToolInvocationError: Failed to call MCP tool
'pollen_robotics_reachy_mini_search_tool__reachy_mini_search_tool_search_web'
from 'pollen_robotics_reachy_mini_search_tool': unhandled errors in a
TaskGroup (1 sub-exception)
ValueError: not enough values to unpack (expected 3, got 2)
```

`mcp_client.py:365`:

```python
read_stream, write_stream, _ = transport
```

`mcp/client/streamable_http.py:705` yields two values:

```python
yield read_stream, write_stream
```

The third value is already discarded, so the fix is:

```python
read_stream, write_stream = transport
```

## Bug 2 — timedelta passed where seconds are expected

After fixing bug 1:

```
TypeError: unsupported operand type(s) for +: 'float' and 'datetime.timedelta'
```

`mcp_client.py:309`:

```python
read_timeout_seconds=timedelta(seconds=self.server.tool_timeout_s),
```

mcp 2.0 expects float seconds:

```python
read_timeout_seconds=self.server.tool_timeout_s,
```

## Note

The call site appears to have been partially migrated to 2.x already — the
import at line 136 uses the 2.x name `streamable_http_client` (1.x exported
`streamablehttp_client`), and line 364 passes the 2.x `http_client=` kwarg.
The two lines above were missed.

Worth also adding an upper bound on the `mcp` dependency so installs don't
silently resolve to a major version the code hasn't been migrated for.

## Verification

Both fixes applied locally; the search and weather tools then work normally.

## Possibly unrelated

The invoked tool name has a doubled prefix
(`..._search_tool__reachy_mini_search_tool_search_web`) while the realtime
layer logs it as `..._search_tool__search_web`.

olaservo:
## Summary

Fixes #524.

`pyproject.toml` declares `mcp>=1.27.1` with no upper bound. `uv.lock` pins 1.28.1, but the daemon installs apps with `uv pip install` (`reachy_mini/apps/sources/local_common_venv.py`), which ignores the lock and re-resolves to mcp 2.0.0.

v2 renamed the model fields to snake_case and kept the wire names as pydantic aliases. Construction still works, so only attribute reads changed.

This pins `mcp==2.0.0` (plus `mcp-types`, which `mcp` exact-pins) and migrates the client.

Note that `Client`'s default `mode="auto"` probes `server/discover` and falls back to the `initialize` handshake, so one client still speaks both protocol revisions.

## Tests

It appeared that `tests/test_mcp_client_integration.py` was being skipped. It now runs against a real in-process `MCPServer` over uvicorn, exercising the transport, discovery, invocation and the tool timeout.

I also added a pagination regression test for the `nextCursor` rename, since that one fails silently and a small tool list hides it.

Verified against a real `mcp==2.0.0` env: `ruff check .`, `ruff format --check .`, `mypy` (45 files, no issues), `pytest tests/` → **310 passed, 0 skipped**. Swapping the current `mcp_client.py` back in fails 7 of the 10 MCP tests, including both integration tests.

Not yet run against a physical robot with a remote MCP Space.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [x] `.env.example` updated (if new config vars were added) — n/a, no new config vars

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,9 @@ dependencies = [
     #Reachy mini
     "reachy_mini_dances_library",
     "reachy-mini>=1.10.0rc2",
-    "mcp>=1.27.1",
+    "mcp==2.0.0",
+    "mcp-types==2.0.0",
+    "httpx2>=2.5.0",
 ]
 
 [dependency-groups]
--- a/src/reachy_mini_conversation_app/mcp_client.py
+++ b/src/reachy_mini_conversation_app/mcp_client.py
@@ -8,16 +8,15 @@ downloading third-party Python code.
 from __future__ import annotations
 import re
 from typing import TYPE_CHECKING, Any, Mapping, Sequence, AsyncIterator
-from datetime import timedelta
 from contextlib import asynccontextmanager
 from dataclasses import field, dataclass
 from urllib.parse import urlparse
 
 
 if TYPE_CHECKING:
-    from mcp import ClientSession
-    from mcp.types import Tool as McpTool
-    from mcp.types import CallToolResult as McpCallToolResult
+    from mcp import Client
+    from mcp_types import Tool as McpTool
+    from mcp_types import CallToolResult as McpCallToolResult
 
 
 _NAME_SEGMENT_PATTERN = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
@@ -110,9 +109,21 @@ def _join_text_content(content_blocks: list[dict[str, Any]]) -> str | None:
     return "\n\n".join(text_parts)
 
 
+def _is_timeout_mcp_error(exc: BaseException) -> bool:
+    """Whether this is the SDK's request-timeout error (JSON-RPC -32001)."""
+    try:
+        from mcp import MCPError
+        from mcp_types import REQUEST_TIMEOUT
+    except ImportError:
+        return False
+    return isinstance(exc, MCPError) and getattr(exc, "code", None) == REQUEST_TIMEOUT
+
+
 def _exception_contains_timeout(exc: BaseException) -> bool:
     timeout_exception = _httpx_timeout_exception_type()
     if isinstance(exc, timeout_exception):
+        return True
+    if _is_timeout_mcp_error(exc):
         return True
 
     if "timed out" in str(exc).lower() or "deadline exceeded" in str(exc).lower():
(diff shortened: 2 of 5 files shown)


andimarafioti:
it seems like @alozowski already addressed this by setting the upper bound for mcp to "<2.0". She also released a new version of the conv app (v1.0.1). That would mean that this PR ins't needed, can you check it?

andimarafioti (approved):
LGTM :)

olaservo:
Hi, wanted to respond here too since I had just mentioned this on the Discord thread.  Even though pinning the version fixes the version mismatch issue, I think it would still make sense to make this update so that Reachy can support the latest MCP spec version (which came along with the v2 sdk).

There is also an issue open to support custom servers, so it should make that PR a little smaller, since custom servers would ideally be able to communicate using the latest version of the protocol.

andimarafioti:
Sorry, I missed the discord thread @olaservo , could you tell me what v2 improves? And is it backwards compatible?

olaservo:
Hi @andimarafioti , good questions!

mcp 2.x supports the 2026-07-28 MCP protocol revision (including stateless requests, `server/discover`, the newer subscription and multi-round-trip request shapes, etc.). 1.x went into maintenance mode when 2.0.0 shipped on 2026-07-28 and only gets security fixes now. 

The practical gain today is that tool Spaces and servers built on mcp 2 can use the new revision and the app talks to them natively instead of through the compatibility handshake. The other thing the PR does is fix the MCP integration test, which was being skipped, so it now runs against a real in-process server and would have caught the 1.x/2.0 break.

Re: backwards compatibility, short answer is yes.  The v2 `Client` defaults to `mode="auto"` which probes `server/discover` and falls back to the `initialize` handshake, so the same client works against 2025-era servers and 2026-07-28 servers. Nothing changes for users pointing the app at an existing tool Space.

I also loosened the pin. I had `mcp==2.0.0` when I opened this and PyPI is at 2.2.0 now, so it is `mcp>=2.0,<3` (and the same for `mcp-types`) with the lock on 2.2.0, so the daemon's `uv pip install` should now track 2.x without another break. Checks pass on 2.2.0, and the branch is merged up to current main.

Issue #434, opened 2026-06-25
Support configuring custom MCP servers


olaservo:
Today Reachy only supports using public Gradio-hosted MCP servers.  Allowing Reachy to use other types of MCP servers has been very useful in my forked version of the app.  Happy to open a PR for this since I've already got a version working locally.

alozowski:
Hello @olaservo, 

That's true, feel free to open a PR and we will iterate on it together 👍🏻

PR #535, opened 2026-08-15
Keep a corrupt tool-Spaces manifest from taking down the tool registry

### LINKED ISSUE #534: A corrupt `installed_tool_spaces.json` takes down the whole tool registry, not just the Space tools
## Summary

`_resolve_remote_tools` in `tools/core_tools.py` reads the manifest without a guard:

```python
for installed_space in read_installed_tool_spaces(instance_path).spaces:
```

Anything the reader raises propagates through `initialize_tools`, so a damaged manifest leaves the app with no tools at all.

## Reproduction

With a profile that enables a Space tool, write a manifest whose envelope is well-formed but whose entry trips the URL validator:

```json
{"version": 2, "spaces": [{"slug": "example/search-tool", "alias": "example_search_tool",
  "mcp_url": "https://attacker.example/gradio_api/mcp/", "private": true, "tools": []}]}
```

`initialize_tools()` then raises `ValueError: Invalid Hugging Face Space MCP URL ...` instead of starting with the built-in tools.

## Suggested fix

Catch both types around the manifest read, log, and return no remote tools.  This is the same degradation the loop already applies to a Space whose cached tools have gone missing.

olaservo:
## Summary

Fixes #534 

## Tests

Added `test_initialize_tools_survives_a_corrupt_installed_spaces_manifest`, which writes a manifest whose envelope is valid but whose entry trips the URL validator, then asserts `initialize_tools()` completes and the built-in tools are still registered. Against `main` it fails with the raw `ValueError` escaping tool initialization.

`ruff check .`, `ruff format --check .`, `mypy` (45 files, no issues), `pytest tests/` → 308 passed.

## Category
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD
- [ ] Other

## Pre-merge checklist
- [x] `.env.example` updated (if new config vars were added) — n/a, no new config vars

## Notes

Found while working on custom MCP servers (#434), which adds a second manifest with the same failure mode. Splitting it out because it stands alone and applies to `main` as it is today.

### DIFF
--- a/src/reachy_mini_conversation_app/tools/core_tools.py
+++ b/src/reachy_mini_conversation_app/tools/core_tools.py
@@ -321,8 +321,16 @@ def _read_profile_tool_names(instance_path: str | Path | None) -> list[str]:
 
 def _resolve_remote_tools(tool_names: list[str], instance_path: str | Path | None) -> list[RemoteMcpTool]:
     """Build Space tools enabled by the active profile from the cached install manifest, without any network calls."""
+    try:
+        manifest = read_installed_tool_spaces(instance_path)
+    except (RuntimeError, ValueError) as exc:
+        # Skip the installed Spaces rather than take the whole registry down.
+        # ValueError too: validate_space_slug and validate_space_mcp_url raise it.
+        logger.error("Skipping installed tool Spaces: %s", exc)
+        return []
+
     remote_tools: list[RemoteMcpTool] = []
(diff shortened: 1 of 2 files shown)


olaservo:
Didn't use the right branch naming convention so re-opend with a re-named branch here: https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/536

PR #545, opened 2026-08-30
feat: make the instance path configurable via REACHY_MINI_INSTANCE_PATH

### LINKED ISSUE #544: Long-term memory is stored inside site-packages and is destroyed by reinstalling the app
### Summary

Long-term memory is written **inside the installed package**, so reinstalling or updating
the app deletes it. The location also changes depending on how the app is launched, so the
same user on the same machine can get two different memory stores.

### Where it lands

`ReachyMiniApp._get_instance_path()` (SDK) returns the app module's own `__file__`, and
`memory_path_for_instance()` joins `memory.v1.json` onto it:

```python
# reachy_mini/apps/app.py
def _get_instance_path(self) -> Path:
    mod = importlib.import_module(type(self).__module__)
    return Path(mod.__file__).resolve()
```

```python
# reachy_mini_conversation_app/memory.py
def memory_path_for_instance(instance_path=None) -> Path:
    if instance_path is not None:
        return Path(instance_path).expanduser() / MEMORY_FILENAME
    data_home = os.getenv("XDG_DATA_HOME")
    data_root = Path(data_home).expanduser() if data_home else Path.home() / ".local" / "share"
    return data_root / "reachy_mini_conversation_app" / MEMORY_FILENAME
```

The `instance_path is None` branch does the right thing — an XDG data directory. It is just
never taken when the daemon launches the app.

| Launch method | Memory file |
| --- | --- |
| Daemon (`POST /api/apps/start-app/...`) | `<venv>/lib/python3.12/site-packages/reachy_mini_conversation_app/memory.v1.json` |
| Standalone (`reachy-mini-conversation-app`) | `~/.local/share/reachy_mini_conversation_app/memory.v1.json` |

### Impact

**Data loss.** Reinstalling or updating the app rewrites that directory. Anything the user
asked Reachy to remember is gone, silently, with no warning that a destructive step is
about to happen.

**Split memory.** Start the app the other way and it has amnesia — same machine, same user,
no indication why. Facts saved in one mode are invisible in the other.

**Discoverability.** Users cannot reasonably find, back up, or delete their own stored
personal data. I went looking for mine and missed it on the first pass, because
`site-packages` is not where anyone expects personal data to live. That matters: `remember`
is documented to store names, preferences, hobbies, important people and plans.

### Reproduce

1. Start the daemon, start the conversation app through it.
2. Tell Reachy something to remember — for example your name.
3. `find . -name memory.v1.json` inside the apps venv — the file is inside the package.
4. Reinstall the app. The file is gone.

Observed on app 1.0.1 with `reachy-mini` 1.10.0, using the **default hosted Hugging Face
backend** — this is not specific to any LLM backend, since the memory path never consults
it. My own store held two facts, the older written 13 days before I had changed anything
about my setup.

### Suggested fix

Make the instance path configurable, matching the pattern the app already uses for
`REACHY_MINI_EXTERNAL_TOOLS_DIRECTORY` and `REACHY_MINI_EXTERNAL_PROFILES_DIRECTORY`:

```
REACHY_MINI_INSTANCE_PATH=~/reachy-mini-data
```

Falling back to current behaviour when unset keeps it non-breaking, and gives users a
single visible directory for memory, profiles and settings that survives reinstalls.

A larger fix would be to default to the XDG path — the branch already written in
`memory_path_for_instance` — and treat `instance_path` as packaged-asset lookup only. That
changes behaviour for existing installs, so it likely wants a migration step.

Happy to send a PR for the env-var version.

lijunzh:
Fixes #544.

### Problem

Long-term memory, profiles and settings are written to the app's instance path, which
resolves to the **installed package directory** when the daemon launches the app:

```python
instance_path = self._get_instance_path().parent   # -> site-packages/reachy_mini_conversation_app/
```

Two consequences:

- Reinstalling or updating the app deletes whatever the user asked Reachy to remember.
- The data is in `site-packages`, where users cannot reasonably find, back up or delete
  their own stored personal information — and `remember` is documented to store names,
  preferences, hobbies, important people and plans.

Launching standalone instead takes the `instance_path is None` branch and lands in an XDG
directory, so the same user on the same machine can end up with two different memory
stores depending on how the app was started.

### Change

`REACHY_MINI_INSTANCE_PATH` overrides the instance directory. Unset, behaviour is
identical to today, so existing installs are unaffected.

```bash
REACHY_MINI_INSTANCE_PATH=~/reachy-mini-data
```

This matches the pattern the app already uses for `REACHY_MINI_EXTERNAL_TOOLS_DIRECTORY`
and `REACHY_MINI_EXTERNAL_PROFILES_DIRECTORY`, so it should feel familiar rather than new
surface.

An unusable value warns and falls back to the previous path instead of failing to start:
a bad env var should not stop the robot from talking.

### Verified

| case | result |
| --- | --- |
| unset | returns the package directory, unchanged |
| valid path | used, and created if missing |
| `~/...` | expanded |
| unusable (`/dev/null/nope`) | warns, falls back, no crash |

`ruff check` and `ruff format --check` pass on the changed file.

### Not included

A larger fix would default to the XDG path — the branch already present in
`memory_path_for_instance` — and treat `instance_path` as packaged-asset lookup only. That
changes behaviour for existing installs and would want a migration step, so I kept this
opt-in and non-breaking. Happy to follow up if you would prefer that direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ Copy `.env.example` to `.env` when you want to point Hugging Face at your own lo
 | `HF_REALTIME_WS_URL` | Direct websocket endpoint for your own Hugging Face backend. Accepts either a base URL like `ws://127.0.0.1:8765/v1` or the full websocket URL `ws://127.0.0.1:8765/v1/realtime`. Used when `HF_REALTIME_CONNECTION_MODE=local`. |
 | `HF_TOKEN` | Optional token for Hugging Face access. Local endpoints receive only this explicitly configured token. |
 | `REACHY_MINI_APP_TIMEOUT_MINUTES` | Minutes of inactivity before Reachy goes to sleep and the app stops. Defaults to `1440` (one day); set to `0` to disable. |
+| `REACHY_MINI_INSTANCE_PATH` | Directory for this instance's data: long-term memory, profiles and settings. Defaults to the installed package directory, where an app reinstall deletes it. Set this to a directory you own to keep that data across updates. |
 
 ### Hugging Face Connection Modes
 
--- a/src/reachy_mini_conversation_app/main.py
+++ b/src/reachy_mini_conversation_app/main.py
@@ -1,6 +1,7 @@
 """Entrypoint for the Reachy Mini conversation app."""
 
 from __future__ import annotations
+import os
 import sys
 import time
 import asyncio
@@ -346,6 +347,36 @@ def run(
         logger.info("Shutdown complete.")
 
 
+def _resolve_instance_path(default: Path) -> Path:
+    """Return the instance directory, honouring REACHY_MINI_INSTANCE_PATH.
+
+    ``default`` is the installed package directory. Writing user data there
+    means an app reinstall deletes it, and it is not somewhere a user can
+    reasonably find or back up their own stored facts. Setting the environment
+    variable puts memory, profiles and settings in a directory the user owns.
+
+    Unset keeps the previous behaviour, so existing installs are unaffected.
+    """
+    override = os.getenv("REACHY_MINI_INSTANCE_PATH")
+    if not override:
+        return default
+
+    path = Path(override).expanduser()
+    try:
+        path.mkdir(parents=True, exist_ok=True)
+    except OSError as exc:
+        logging.getLogger(__name__).warning(
+            "REACHY_MINI_INSTANCE_PATH=%s is unusable (%s); falling back to %s",
+            override,
+            exc,
+            default,
+        )
+        return default
+
+    logging.getLogger(__name__).info("Using instance path %s", path)
+    return path
+
+
 class ReachyMiniConversationApp(ReachyMiniApp):  # type: ignore[misc]
     """Reachy Mini Apps entry point for the conversation app."""
 
@@ -358,7 +389,7 @@ class ReachyMiniConversationApp(ReachyMiniApp):  # type: ignore[misc]
 
         args, _ = parse_args()
 
-        instance_path = self._get_instance_path().parent
+        instance_path = _resolve_instance_path(self._get_instance_path().parent)
         run(
             args,
             robot=reachy_mini,


FabienDanieau (changes requested):
Thanks for the precise diagnosis. Requesting changes because the fix as written will not reach the robots that have the problem, and because one premise in #544 did not reproduce.

**The default still runs on every shipped robot.** The daemon starts apps with a copy of its own environment and nothing in the OS image sets `REACHY_MINI_INSTANCE_PATH`. It cannot come from the app's `.env` either: that file lives inside the instance directory and `run()` loads it after `instance_path` is resolved (`main.py` L140-148). Checked on a Wireless unit: the live daemon environment has no such variable, and the installed app resolves memory to `site-packages` exactly as your table says.

**Reinstall does not delete the file, though.** On the robot's own Python and pip (the daemon's installer; `uv` is not present there), `pip uninstall -y` + install, `--force-reinstall`, and a version upgrade all leave a stray `memory.v1.json` in the package directory. Your loss most likely came from a recreated venv (reflash, factory reset). Worth stating the risk that way in the README row.

**Proposal: default to the daemon's config directory, keep the env var as an override.** The daemon persists its own state under `platformdirs.user_config_dir("reachy_mini")`, `~/.config/reachy_mini/` on the robot (`daemon/startup_app_config.py`, `utils/robot_name.py`). Default the instance directory to `<that>/conversation_app/`. `platformdirs` is a declared dependency of `reachy-mini`; declare it here and `uv lock`.

**Migration fits in the PR.** If the new directory holds none of `memory.v1.json`, `.env`, `startup_settings.json`, `profile_toolsets.json`, `user_personalities/`, and the legacy package directory holds any, copy them once and log it. That removes the reason to stay opt-in.

Smaller: no test (`monkeypatch.setenv` + `tmp_path` covers it); `.env.example` needs the variable; use the module logger; one-line docstring per AGENTS.md.

Happy to pair on the migration, or we can pick it up with credit.

PR #552, opened 2026-09-16
Settings over JSON-RPC, stored in the user config dir instead of .env

### LINKED ISSUE #532: A settings value containing a line break adds an extra assignment to the instance .env
## Summary

`LocalStream._persist_env_values` writes `f"{env_name}={value}"` with no quoting, which can cause issues in the following scenarios:

**A line break ends the record and assigns the rest to the following key:**

`_rpc_backend_config` validates the Hugging Face host against `://`, `/`, `?` and `#`, but not against whitespace or newlines. A host containing a line break reaches `_persist_env_values` and is written verbatim:

```
HF_REALTIME_WS_URL=ws://host
API_KEY=foo
```

python-dotenv then parses `API_KEY` as a real key. Reproduced end to end against `main`.

**`${...}` is another case:** 

python-dotenv interpolates it in every quoting style with no escape syntax, so a value containing `${HOME}` is stored expanded rather than literal.

## Impact

Any value the settings UI persists (such as tokens) can end up altered between what the user typed and what the app reads back at the next start. The line-break case additionally lets one settings field write an unrelated key into the `.env`. 

Note that this requires access to the settings API, so it is not remotely exploitable on its own.

## Suggested fix

Quote values containing whitespace, `#`, a quote or a backslash.

### LINKED ISSUE #544: Long-term memory is stored inside site-packages and is destroyed by reinstalling the app

FabienDanieau:
Fixes #544. Closes #532.

## Problem

The instance path is derived from the package's own `__file__`, so long-term memory and user personalities live inside site-packages and are destroyed by a reinstall or update (#544). UI settings are appended to a `.env` in that same place with no quoting, so a value containing a line break spills into the next key (#532). And a remote client such as the mobile app has no way to read or change memory, transcription language or camera use: the only surface for those was the local settings page.

## Solution

- The instance path resolves to `platformdirs.user_config_dir("reachy_mini")`, the same directory the daemon uses, and existing data (memory, user personalities, the instance `.env`) is migrated there once.
- UI settings persist in a `settings.json` under that path instead of the `.env`; the `.env` is only read, never written.
- New JSON-RPC methods on `/rpc`: `memory.list/forget/clear/set_enabled`, `language.get/set`, `vision.get/set`. `conversation.status` also reports personality, voice, language, memory and vision so a client needs one round-trip on connect.
- `memory.clear` refreshes the live session's instructions when a backend is connected, so the model forgets at once instead of at the next restart.
- A real `memory.enabled` setting: disabled, remembered facts stay out of the prompt and the remember/forget tools decline.

Tested on a Wireless robot with the mobile app branch (pollen-robotics/reachy_mini_mobile_app#63) driving it over the daemon relay.


FabienDanieau:
Needed by pollen-robotics/reachy_mini_mobile_app#96, which uses these methods for the settings panel.

PR #553, opened 2026-09-17
Ship the knowledge tool Space as a default

FabienDanieau:
The knowledge tool Space (`pollen-robotics/reachy-mini-knowledge-tool`) answers questions about Reachy Mini, MicroDuck and Pollen Robotics, but users have to install it by hand — a fresh robot can't answer "what is a Reachy Mini?".

Seed it in `PREINSTALLED_TOOL_SPACE_SPECS` next to the search, time and weather Spaces, and enable its two tools (`get_info`, `search_docs`) in the default profile. Tool names, descriptions and schemas come from the Space's live MCP schema, so startup still needs no Hugging Face discovery.

Tested on a robot: seeded manifest exposes both tools and the assistant calls them.


RemiFabre (changes requested):
I tested this by asking questions to the robot. Everything worked except one question (about troubleshooting on Reachy Mini) that remained silence, after a while I asked what happened and the robot told me that the request had timed out. imo:
- There should be a small sound effect (happy dong) every time an external call is made, it clarifies what's happening to the user and it makes it more natural to wait for an answer (even if the delays are already relatively short)
- In case of failure, the robot should never just stay silent. Maybe another (different) sound effect (sad dong) + the robot commenting the failure
- Also the manifest (JSON) thing is a bit annoying. My understanding is that if a user had already installed an MCP tool, and we push an update that sets "knowledge-tool" as a default MCP tool, that user will NOT have it ON, correct? Because there would be a conflict between our default manifest and theirs. Maybe we could merge them on upgrades, so that simple additive tools like this are available by default? Otherwise I feel like 99% of our users will never realize there is a new capability available.

Other than that this is a great improvement!

FabienDanieau:
> I tested this by asking questions to the robot. Everything worked except one question (about troubleshooting on Reachy Mini) that remained silence, after a while I asked what happened and the robot told me that the request had timed out. 

It seems more an issue on the MCP tooling right? I feel that we should open an issue addressing this timeout problem rather than fixing it in this PR.

PR #555, opened 2026-09-17
Let the assistant inspect and control the robot itself

### LINKED ISSUE #554: Let the assistant inspect and control the robot itself
Reachy can move and express itself, but it cannot answer anything about its own state. There is no way to ask it to turn its volume up, what version it runs, what its IP is, which account it is signed into, what apps are installed, or whether its head is straight.

The daemon already exposes all of this over its REST API; the app just never reaches for it (the one existing call, stopping the current app, has its HTTP handling inlined in `app_lifecycle`).

Add the tools for it, on a shared daemon REST client.

FabienDanieau:
Closes #554

Reachy could not answer anything about its own state: turn the volume up, what version it runs, what its IP is, which account it is signed into, what apps are installed, whether its head is straight. The daemon exposes all of it over REST; the app never reached for it.

Two tools, on a shared `daemon_api.daemon_request` factored out of the one HTTP call that was already inlined in `app_lifecycle`:

- `volume_control` — speaker and microphone, reads when no level is given.
- `robot_status` — one `topic` enum (`name`, `software`, `wifi`, `account`, `imu`, `apps`) rather than six near-identical zero-argument tools, to keep the function-calling spec small.

The `imu` topic reports a direction, not raw vectors: gravity in the head frame gives the tilt, and both signs were checked against `present_head_pose` on a wireless unit (28.6 vs 28.4 deg nose-down, 44.4 vs 43.5 deg tilted right).

All seven readings were verified against a live daemon on a Wireless unit running 1.11.0rc1. Only the default profile enables the tools.


RemiFabre (approved):
We should've had this a long time ago, thank you! I tested this by asking questions to the robot for each of the new introspection tools. Everything worked.

